PR #4, opened 2018-05-07
Create LICENSE

stsp:
freedos has GPL2+ license.
AFAICS there is no problem to cover fdpp
under GPL3+.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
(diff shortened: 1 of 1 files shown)


stsp:
@bartoldeman Could you please confirm this
is OK to do?

bartoldeman:
yes the freedos kernel allows it ("at your option" wording in many places). However you will need to adjust some files such as docs/copying too!

some other components of freedos are not GPL though, see http://www.ibiblio.org/pub/micro/pc-stuff/freedos/files/repositories/1.2/pkg-html/group-base.html

stsp:
OK, thanks for info!
This is a github-generated PR, so I'll
probably apply it as-is, and then do
some manual adjustments as you suggest.

PR #17, opened 2018-07-21
Debug02

andrewbird:
Capture the DOS debug output and route to dosemu.

### DIFF
--- a/fdpp/thunks.cc
+++ b/fdpp/thunks.cc
@@ -357,7 +357,8 @@ void fdlogvprintf(const char *format, va_list vl)
     fdpp->print(1, format, vl);
 }
 
-void fdprintf(const char *format, ...)
+
+static void _fdprintf(const char *format, ...)
 {
     va_list vl;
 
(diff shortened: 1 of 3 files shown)


fdpp/thunks.cc
```diff
@@ -366,7 +367,19 @@ void fdprintf(const char *format, ...)
     va_end(vl);
 }
 
-void fdlogprintf(const char *format, ...)
+void fdprintf(const char *format, ...)
```
stsp: I don't think we need to re-do fdprintf()
that way. fdprintf() is for our own debugging,
say, in thunks.cc, so it should be compatible
with dosemu.
I meant the following:
```
You should define the new function that
will to _sprintf() into the buffer, and then
pass it to fdprintf() already with things
like %P expanded.
```

Also are you sure you need to re-define
`DebugPrintf()`? You were saying this:
`After getting tn_printf() in truename to print to dosemu I find that dosemu crashes.`
Do you really need to redefine anything
more that `tn_printf()`? `DebugPrintf()`
prints to dosemu terminal anyway, or
do you want into the log?

stsp:
Basically we need to find out why is this needed.
1. This is an incompatible change that will add
troubles to whoever wants to restore the old
compilation, because in old compilation there
is no fdpp API to print.
2. The debug output goes to the dosemu screen
anyway even now, so what is to change? Do
you want it to go to dosemu log?

stsp:
Didn't mean to close.
What _is_ the problem now, is that I used
several `snprintf()` calls that just shouldn't
be there. If you replace it with `_snprintf()`,
that would be good.

fdpp/thunks.cc
andrewbird: At the time %P was being passed through to dosemu, which was ignored as it doesn't know how to deal with it, so the following parameters didn't match the format passed.  I believe that caused the crash.

stsp: Yes, but why to re-define `fdprintf()` for this?
In particular, why to even want to print debug
via fdpp API, if they go to dosemu screen anyway?

hdr/debug.h
```diff
@@ -49,7 +49,7 @@
 #ifdef DEBUG_PRINT_COMPORT
 #define DebugPrintf(x) dbgc_printf x
 #else
-#define DebugPrintf(x) _printf x
+#define DebugPrintf(x) fdprintf x
```
stsp: It should be done as a new backend,
like `DEBUG_PRINT_DOSEMU`.
But as it actually goes to dosemu anyway,
what problem does this solve?

Issue #36, opened 2018-10-17
audit MK_FAR_STR_SCP uses


stsp:
As pointed in https://github.com/stsp/fdpp/issues/34#issuecomment-430416133,
`FcbNameInit()` didn't work properly with this macro.
Another wrong use is newstuff.c:339.
The rest seems safe.

Issue #35, opened 2018-10-17
convert fcb code to flat pointers


stsp:
As shown in #34, the `FcbRename()` function's data
flow is too difficult for fdpp to compile properly.
Hacks like this:
https://github.com/stsp/fdpp/commit/734847387e9aadaf464a78369378aeaba22cd20e
are not plugging all the holes.
It may be good to convert it to flat pointers.
Or maybe think up something else, but not
multiple the nasty hacks.

stsp:
Looks quite impossible because of those
`dta` assignments between operations...
So as it now works, consider fixed.

PR #37, opened 2018-10-17
Fcb rename 01

andrewbird:


stsp:
Thanks for putting things together.
But I am inclined to treat this as a bug reports,
rather than the finished fixes.
The pointer hack can be applied as is, as it
just completes the initial hack that is already
applied. I'll revert them both if/when I find the
better solution.
The 12 (or 13) one was clearly an example.
We should either pass the buffer size separately,
or maybe just reference the `SecPathName` buffer
directly, rather than passing the pointers to it?
As for shadowing, could you please clarify what
exactly does it fix? Shadowing should work properly,
how it happens that the return value is different?

stsp:
I opened #35 and #36 to think more about these problems.

andrewbird:
Okay I'll drop the last two, you can fixup, call them what you like as you wish. I'll fix the commit message in the first later. Incidentally, the shadowing worked too well there, when the inner result went out of scope the value set was obviously lost and the function always returned FCB_SUCCESS, hence the problem reported to freedos.

kernel/fcbfns.c
```diff
@@ -530,7 +530,7 @@ UBYTE FcbRename(xfcb FAR * lpXfcb)
   {
     dmatch Dmatch;
     dmatch FAR * Dmatch_p;
-    COUNT result;
+    COUNT _result;
```
stsp: Why not to just remove this line and forget?

stsp:
I applied the simpler fix to the shadowing - just removed it. :)

stsp:
Applied the third one by hands.

andrewbird:
> I applied the simpler fix to the shadowing - just removed it. :)

Simpler but doesn't work, returns failure on success. The types of the two result vars are different, you are now testing an unsigned byte var for < 0.
Just to be sure nothing else changed I backed out your simple fix, reapplied mine and it worked as before.

stsp:
Ah, sorry about that.
Only after debugging it myself, I now see
your point, and why it should be the different
var, not the same one.
I rebased my patch out. Please update your
PR and I'll apply.

stsp:
The comment why another var is needed,
would be good. Also I suggest the name `rc`
for new var or alike, to avoid confusion.

stsp:
Applied, thanks.

andrewbird:
> The comment why another var is needed,
would be good. Also I suggest the name rc
for new var or alike, to avoid confusion.

Thanks for the advice, I redid the patch for FreeDOS and it looks lots better with 'rc'.

PR #42, opened 2018-11-20
FCB: Rename should support asterisk wildcards

andrewbird:
Since DOS 3.0 the FCB rename function int21/17 supports asterisk
wildcards. Add a function to expand the '*' into '?' to fill the
remaining buffer.

This changes FDPP to behave the same as PC-DOS, MS-DOS and DR-DOS for both FAT and redirector based filesystems.

Style questions:
- I may need to add something similar for int21/13 Delete, do you prefer squashing with this into a single patch?
- Do you prefer four calls to `FcbExpandAsterisks()`,  one with duplication inside the function, or two calls to function that handles filename and extension?

### DIFF
--- a/kernel/fcbfns.c
+++ b/kernel/fcbfns.c
@@ -502,12 +502,38 @@ UBYTE FcbDelete(xfcb FAR * lpXfcb)
   return result;
 }
 
+static void FcbExpandAsterisks(char *p)
+{
+  int i, afound;
+
+  // Filename
+  for (i = 0, afound = 0; i < 8; i++) {
+    if (p[i] == '*' || afound) {
+      p[i] = '?';
+      afound = 1;
+    }
+  }
+
+  // Extension
+  for (i = 8, afound = 0; i < 11; i++) {
+    if (p[i] == '*' || afound) {
+      p[i] = '?';
+      afound = 1;
+    }
+  }
+}
+
 UBYTE FcbRename(xfcb FAR * lpXfcb)
 {
   rfcb FAR *lpRenameFcb;
   COUNT FcbDrive;
   UBYTE result = FCB_SUCCESS;
   void FAR *lpOldDta = dta;
+
+  /* DOS 3.0+ allows * wildcards for rename */
+  rfcb FAR *trfcb = (rfcb FAR *)lpXfcb;
+  FcbExpandAsterisks(trfcb->renOldName);
+  FcbExpandAsterisks(trfcb->renNewName);
 
   /* Build a traditional DOS file name                            */
   lpRenameFcb = (rfcb FAR *) CommonFcbInit(lpXfcb, SecPathName, &FcbDrive);


andrewbird:
I added a tweak to see if you prefer the function this way?

kernel/fcbfns.c
```diff
@@ -502,13 +502,32 @@ UBYTE FcbDelete(xfcb FAR * lpXfcb)
   return result;
 }
 
+static void FcbExpandAsterisks(char *p, int len)
+{
+  int i, afound;
+
+  for (i = 0, afound = 0; i < len; i++) {
+    if (p[i] == '*' || afound) {
```
stsp: But does this properly handle, say, `a*.b*`?
Or `a*.bat*`?
I mean, you don't seem to be handling dot, which
may have a separate wildcard.
Also not sure if the cases like `*a.*t` are valid?
I think the code should be `if (p[i] == '*' || (afound && p[i] == ' ')) {`
or something like that.

stsp:
> Do you prefer four calls to FcbExpandAsterisks(), one with duplication inside the function,
> or two calls to function that handles filename and extension?

If you pass start and length, then why you need more
than one func?
By the way, what is it, a freedos bug? It wasn't there?

kernel/fcbfns.c
andrewbird: The page I linked to in the commit explains better, but essentially it's very simple and not rich like unix wildcards.
so
`*a.b*` ==> `???????.b??`
`a*.bat*` is illegal
`*a.*t` -> `????????.???`

andrewbird:
I prefer the second commit now, as it reads better in the caller and will be useful for FCB delete too.

Yep it's a freedos bug(or unimplemented feature), but I gave up reporting things as nobody is interested there.

stsp:
OK, please squash then.
Its a bit strange to add and change the
code in a subsequent commits.

kernel/fcbfns.c
andrewbird: There's no dot as the input is an FCB name (8+3 chars, no dot, no \\0).

kernel/fcbfns.c
```diff
@@ -502,13 +502,39 @@ UBYTE FcbDelete(xfcb FAR * lpXfcb)
...
+
 UBYTE FcbRename(xfcb FAR * lpXfcb)
 {
   rfcb FAR *lpRenameFcb;
   COUNT FcbDrive;
   UBYTE result = FCB_SUCCESS;
   void FAR *lpOldDta = dta;
 
+  /* DOS 3.0+ allows * wildcards for rename */
+  rfcb FAR *trfcb = (rfcb FAR *)lpXfcb;
```
stsp: Why is such cast?
Are these structs compatible?

andrewbird: Actually the pointer passed into FcbRename should be an `rfcb *`, but it's pushed through a generic parsing routine to get the first filename extracted as they are initially compatible. Later they cast it back to get to the second part of the structure. All pretty horrid if you ask me.

stsp: There seem to be the `GetNameField()` in fcbfns.c,
which does similar to what you do?

andrewbird: Seems to be operating with asciiz strings, we have 11 chars, no dot, and no terminator.

stsp: It seems to be very generic: has `nFieldSize` for
non-asciiz. And more importantly, it is called from
`FcbParseFname()`, which, in turn, is called from
`FcbRename()`!

andrewbird: If that's all working then perhaps it's just the destination pattern that's not being expanded?

stsp: I bet.
Plus in this case you likely need to not only
expand, but also match with the from_name,
so that there no ? in the to_name at the end.

stsp:
So could you please convert this to the bug-report
with a test-case then? I'll take a look when time permits.

andrewbird:
Sure, I lost interest in fixing FreeDOS/FDPP stuff. Here's the test case source, and you'll need to assemble for each test with the fn1,fe1,fn2,fe2 updated accordingly as it doesn't handle command line args.

```asm
.text                                                                           
.code16                                                                         
                                                                                
    .globl  _start16                                                            
_start16:                                                                       
                                                                                
    push    %%cs                                                                
    pop     %%ds                                                                
                                                                                
    movw    $0x1700, %%ax                                                       
    movw    $fcb, %%dx                                                          
    int     $0x21                                                               
                                                                                
    cmpb    $0, %%al                                                            
    je      prsucc                                                              
                                                                                
prfail:                                                                         
    movw    $failmsg, %%dx                                                      
    jmp     1f                                                                  
prsucc:                                                                         
    movw    $succmsg, %%dx                                                      
1:                                                                              
    movb    $0x9, %%ah                                                          
    int     $0x21                                                               
                                                                                
exit:                                                                           
    movb    $0x4c, %%ah                                                         
    int     $0x21                                                               
                                                                                
fcb:                                                                            
    .byte   0       # 0 default drive                                           
fn1:                                                                            
    .ascii  "% -8s"    # 8 bytes                                                
fe1:                                                                            
    .ascii  "% -3s"    # 3 bytes                                                
wk1:                                                                            
    .space  5                                                                   
fn2:                                                                            
    .ascii  "% -8s"    # 8 bytes                                                
fe2:                                                                            
    .ascii  "% -3s"    # 3 bytes                                                
wk2:                                                                            
    .space  16                                                                  
                                                                                
succmsg:                                                                        
    .ascii  "Rename Operation Success$"                                         
failmsg:                                                                        
    .ascii  "Rename Operation Failed$"                                          
```

Test one:
~~~
            fn1 = "*"                                                           
            fe1 = "in"                                                          
            fn2 = "*"                                                           
            fe2 = "out"                                                         
            create the following files "one.in", "two.in", "three.in", "four.in", "five.in", "none.ctl"
  Run test:
            check "one.out, "two.out", "three.out", "four.out", "five.out", "none.ctl" are present
~~~

Test two:
~~~
            fn1 = "a*"                                                          
            fe1 = "*"                                                           
            fn2 = "b*"                                                          
            fe2 = "out"                                                         
            create "aone.in", "atwo.in", "athree.in", "afour.in", "afive.in", "xnone.ctl"
  Run test:
            check "bone.out", "btwo.out", "bthree.out", "bfour.out", "bfive.out", "xnone.ctl" are present
~~~

Test three:
~~~
            fn1 = "abc0??"                                                      
            fe1 = "*"                                                           
            fn2 = "???6*"                                                       
            fe2 = "*"                                                           
            create "abc001.txt", "abc002.txt", "abc003.txt", "abc004.txt", "abc005.txt", "abc010.txt", "xbc007.txt"
Run test:
            check  "abc601.txt", "abc602.txt", "abc603.txt", "abc604.txt", "abc605.txt", "abc610.txt", "xbc007.txt" are present
~~~

Test Four:
~~~
            fn1 = "abc*"                                                        
            fe1 = "htm"                                                         
            fn2 = "*"                                                           
            fe2 = "??"                                                          
            create "abc001.htm", "abc002.htm", "abc003.htm", "abc004.htm", "abc005.htm", "abc010.htm", "xbc007.htm"
Run test:
           check "abc001.ht", "abc002.ht", "abc003.ht", "abc004.ht", "abc005.ht", "abc010.ht", "xbc007.htm" are present
~~~

stsp:
How about attaching binaries?

andrewbird:
Okay, will do later.

stsp:
Preferably in a new ticket then.

stsp:
Not sure why you say "FreeDOS/FDPP
stuff" in this context. In the past you said
you got tired of freedos because its
difficult to build, debug and reports are
ignored. fdpp addresses all of these:
easiest dos to build and debug, in the
whole world (unlikely some other dos
is built that easily or debugged with gdb),
and the tickets are never ignored.
So its a dos of your dream not less,
get with it. :)

andrewbird:
I have a short attention span these days, and if I don't grok the code,  or I find things otherwise awkward I move on. No offence intended regarding FDPP, I'm sure you are right.

stsp:
Well you can note the awkward code into wiki,
so that it can later be cleaned up. I won't hesitate
to change the freedos code now, as the back-merge
is unlikely to happen anyway. But I am not looking
into it frequently enough to notice the awkward things,
so the hints would help.

PR #46, opened 2018-12-24
kernel: fix: kernel_start should point ss at init data, not code

tkchia:


stsp:
Does this have something to do with this?
https://github.com/stsp/fdpp/commit/a21e02f146373cfb352ee35f42a21ad82abdbf72

stsp:
I reverted the previous patch and fixed it
differently. The bug came from freedos linker
script again.
I am not sure however if it fixes the things for
you, as for me these 2 patches produce the
same result. So please see if its still broken.

stsp:
I also tried your patch and it breaks UMB A0
for me again. So what I think is that we have the
different versions of linker, that differently react
on the forgotten SIZEOF bug. So I hope my current
fix would fix it for both of us.

stsp:
In fact, I think the missing SIZEOF didn't lead
to any bug for your (presumably much newer)
ld, but my ld falls apart.
I tried this now with gold, and it asserts!
Any sane linker around? :)

stsp:
So my current conclusion is that there was no
bug (or a very small one) in the freedos linker
script, but we have a bug in my version of ld
and gold. That made me to create the wrong
patch that is now reverted.
With a missing SIZEOF now added, hope both
buggy and sane ld will behave correctly, and gold
got a bug-report filled:
https://sourceware.org/bugzilla/show_bug.cgi?id=24033

stsp:
Hi @tkchia, do you still have the system with
old ld where that problem was reproducible?
Alan Morda in binutils bugzilla suggested to
me another fix:
```
diff --git a/fdpp/kernel.ld b/fdpp/kernel.ld
index 143afe8..be56f98 100644
--- a/fdpp/kernel.ld
+++ b/fdpp/kernel.ld
@@ -74,5 +74,5 @@ SECTIONS
                *(_STACK)
                . = 0x1000;
        }
-    /DISCARD/ : { *(.*) }
+    /DISCARD/ : { *(*) }
   }

```

Of course it have to be applied to the revision
that pre-dates 263514f8, or you won't notice any
difference. And on my newer ld it (expectedly)
also does not produce any difference, but on
yours it should (according to Alan).
So it would be nice if you can test that fix.
If it appears to be also fixing the problem, then
I'll apply it. Unfortunately I can't justify it myself.

You have probably already noticed that I also
am keeping to fix the other bug you reported:
#49 #50. The proper fix appeared to be quite
intrusive.

PR #84, opened 2019-06-30
Travis: Add build

andrewbird:
Simple build check for every commit.

BTW if you change your workflow to always push to a branch, then only after the check succeeds merge it to master, the silly errors will be caught.


stsp:
You mean by creating the PRs to myself?

PR #89, opened 2019-07-14
api: Add support for kernel symbol handling

andrewbird:
1/ Add the ability to return the kernel map file name

2/ By adding a relocation notification function in the API a
previously loaded set of symbols may be moved as the memory is
relocated.

3/ Install the kernel map file alongside the kernel itself.

### DIFF
--- a/fdpp/fdpp.spec.in
+++ b/fdpp/fdpp.spec.in
@@ -45,5 +45,6 @@ make DESTDIR=%{buildroot} PREFIX=%{_prefix} LIBDIR=%{_libdir} install
 %{_includedir}/fdpp/thunks.h
 %{_includedir}/fdpp/bprm.h
 %{_datadir}/fdpp/fdppkrnl.sys
+%{_datadir}/fdpp/fdppkrnl.map
 
 %changelog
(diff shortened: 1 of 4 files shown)


stsp:
The relocations start not at the
beginning of segment. They start
at offset and end at len+offset.

andrewbird:
okay I'll redo.

andrewbird:
So how about `len` i see it's a uint32_t as it's in bytes can it be larger than 0xffff?

andrewbird:
I force pushed the changes, they're trivial here. See the dosemu branch for the maths.

stsp:
Initially I thought about the multi-segment
relocations, where old_seg and new_seg
would only be used to calculate delta, and
len can be large. But in the latest betas
the segments are explicitly compared to
old_seg and replaced with new_seg.
In this scheme it is very unlikely that len
can exceed 0xffff. This solves the problem
with boundary symbols, ie you may have
the end boundary of prev seg and
begin boundary of next seg at the same
address. Distinguishing between them
can only be done by comparing their
segments explicitly, and that leaves no
room for multiseg relocations.

Will look at the patch later.

PR #111, opened 2019-11-10
inthndlr.c: in Int21.43FF, dispatch DosMkRmdir on CL not AH

andrewbird:
As per discussion in https://github.com/PerditionC/fdkernel/pull/33#issuecomment-552095063 here's ecm's fix

### DIFF
--- a/kernel/inthndlr.c
+++ b/kernel/inthndlr.c
@@ -968,7 +968,7 @@ dispatch:
                 case 0x39:
                 /* Dos Remove Directory                                         */
                 case 0x3a:
-                rc = DosMkRmdir(FP_DS_DX, lr.AH);
+                rc = DosMkRmdir(FP_DS_DX, lr.CL);
                 goto short_check;
 
                 /* Dos rename file */


stsp:
I can in principle just do git pull
on their branch if there is nothing
obviously wrong.

andrewbird:
Will that work? I imagined things would be more complicated to transfer stuff from FreeDOS to FDPP so I did `wget http://github.com/xxxxxx.patch` then `git am xxxxxx.patch`.

stsp:
That should work, trying to repair
my pc and check that out.

stsp:
OK, I have obviously underestimated them. :)
In fact they have a lot (I mean, quite a lot, and not
just from tkchia!) of new stuff and the straight-forward
merging is indeed not possible.

stsp:
> so I did wget http://github.com/xxxxxx.patch then git am xxxxxx.patch.

It is better to set them up as another branch
in your local repo. Then you can just do cherry-picks.

stsp:
fdac78c seems to be the last merge point
(141e6373 is a merge commit in fdpp).
```
$ git diff -r fdac78c fdkernel |diffstat
 50 files changed, 1070 insertions(+), 293 deletions(-)
```
Quite a lot, but its not very accurate because
I have ported the elf patches by hands:
```
$ git diff -r d9cfeb8^^ d9cfeb8 |diffstat
 10 files changed, 140 insertions(+), 112 deletions(-)
```
OK, even if we subtract that - still quite a lot.

andrewbird:
> It is better to set them up as another branch
in your local repo. Then you can just do cherry-picks.

Sounds like a good plan!

PR #116, opened 2019-12-07
Travis: Move package add to config

andrewbird:
Now that Travis' Bionic support is ex-beta move the deb package
install to the config instead of using apt-get in a script.


stsp:
Thanks.

PR #119, opened 2019-12-12
fcbfns: Add pointer workaround to FCBDelete [fixes #118]

### LINKED ISSUE #118: FCB Delete broken on FDPP but okay on FreeDOS 1.20
See the [tests](https://travis-ci.com/andrewbird/dosemu2/builds/140839891) running on Travis for my Dosemu2 branch [travis-ci-08](https://github.com/andrewbird/dosemu2/tree/travis-ci-08)

Here's the simple test program [t1.zip](https://github.com/stsp/fdpp/files/3957247/t1.zip) that tries to delete file `testa.bat` in the current directory. On FAT the return value seems to be failure, on MFS the return is success but the files don't seem to be deleted either.

andrewbird:
Similar to FCBRename fix of 5fb6f0bca

Test PP-DOS-GIT FAT FCB file delete missing                             ... ok
Test PP-DOS-GIT FAT FCB file delete simple                              ... ok
Test PP-DOS-GIT FAT FCB file delete wildcard one                        ... ok
Test PP-DOS-GIT FAT FCB file delete wildcard two                        ... ok
Test PP-DOS-GIT FAT FCB file delete wildcard three                      ... ok

Test PP-DOS-GIT MFS FCB file delete missing                             ... ok
Test PP-DOS-GIT MFS FCB file delete simple                              ... ok
Test PP-DOS-GIT MFS FCB file delete wildcard one                        ... ok
Test PP-DOS-GIT MFS FCB file delete wildcard two                        ... ok
Test PP-DOS-GIT MFS FCB file delete wildcard three                      ... ok

### DIFF
--- a/kernel/fcbfns.c
+++ b/kernel/fcbfns.c
@@ -480,10 +480,11 @@ UBYTE FcbDelete(xfcb FAR * lpXfcb)
   else
   {
     int attr = (lpXfcb->xfcb_flag == 0xff ? lpXfcb->xfcb_attrib : D_ALL);
-    dmatch Dmatch;
+    dmatch _Dmatch;
     dmatch FAR * Dmatch_p;
+#define Dmatch (*Dmatch_p)
 
-    Dmatch_p = MK_FAR(Dmatch);
+    Dmatch_p = MK_FAR(_Dmatch);
     dta = Dmatch_p;
     if ((CritErrCode = -DosFindFirst(attr, SecPathName)) != SUCCESS)
     {
@@ -501,6 +502,7 @@ UBYTE FcbDelete(xfcb FAR * lpXfcb)
       }
     }
     while ((CritErrCode = -DosFindNext()) == SUCCESS);
+#undef Dmatch
   }
   dta = lpOldDta;
   return result;
@@ -528,12 +530,13 @@ UBYTE FcbRename(xfcb FAR * lpXfcb)
   }
   else
   {
-    dmatch Dmatch;
+    dmatch _Dmatch;
     dmatch FAR * Dmatch_p;
+#define Dmatch (*Dmatch_p)
     COUNT _result;
 
     wAttr = (lpXfcb->xfcb_flag == 0xff ? lpXfcb->xfcb_attrib : D_ALL);
-    Dmatch_p = MK_FAR(Dmatch);
+    Dmatch_p = MK_FAR(_Dmatch);
     dta = Dmatch_p;
     if ((CritErrCode = -DosFindFirst(wAttr, SecPathName)) != SUCCESS)
     {
@@ -585,6 +588,7 @@ UBYTE FcbRename(xfcb FAR * lpXfcb)
       }
     }
     while ((CritErrCode = -DosFindNext()) == SUCCESS);
+#undef Dmatch
   }
   dta = lpOldDta;
   return result;
--- a/kernel/fcbfns.c
+++ b/kernel/fcbfns.c
@@ -545,7 +545,7 @@ UBYTE FcbRename(xfcb FAR * lpXfcb)
       BYTE loc_szBuffer[2 + FNAME_SIZE + 1 + FEXT_SIZE + 1];
       fcb LocalFcb;
       BYTE *pToName;
-      const char FAR *pFromPattern = Dmatch_p->dm_name;
+      const char FAR *pFromPattern = Dmatch.dm_name;
       const char *pToPattern = buf;
       int i;
       UBYTE mode = 0;
@@ -565,7 +565,7 @@ UBYTE FcbRename(xfcb FAR * lpXfcb)
 
       SecPathName[0] = 'A' + FcbDrive - 1;
       SecPathName[1] = ':';
-      strcpy(&SecPathName[2], Dmatch_p->dm_name);
+      strcpy(&SecPathName[2], Dmatch.dm_name);
       _result = truename(SecPathName, PriPathName, 0);
 
       if (_result < SUCCESS || (_result & IS_DEVICE))
@@ -652,8 +652,8 @@ UBYTE FcbFindFirstNext(xfcb FAR * lpXfcb, BOOL First)
   /* Reconstrct the dirmatch structure from the fcb - doesn't hurt for first */
   Dmatch_ff.dm_drive = lpFcb->fcb_sftno;
 
-  fmemcpy(Dmatch_ff_p->dm_name_pat, lpFcb->fcb_fname, FNAME_SIZE + FEXT_SIZE);
-  DosUpFMem((BYTE FAR *) Dmatch_ff_p->dm_name_pat, FNAME_SIZE + FEXT_SIZE);
+  fmemcpy(Dmatch_ff.dm_name_pat, lpFcb->fcb_fname, FNAME_SIZE + FEXT_SIZE);
+  DosUpFMem((BYTE FAR *) Dmatch_ff.dm_name_pat, FNAME_SIZE + FEXT_SIZE);
 
   Dmatch_ff.dm_attr_srch = wAttr;
   Dmatch_ff.dm_entry = lpFcb->fcb_strtclst;
--- a/kernel/fcbfns.c
+++ b/kernel/fcbfns.c
@@ -653,17 +653,20 @@ UBYTE FcbFindFirstNext(xfcb FAR * lpXfcb, BOOL First)
 
   /* Next initialze local variables by moving them from the fcb   */
   lpFcb = CommonFcbInit(lpXfcb, SecPathName, &FcbDrive);
-  /* Reconstrct the dirmatch structure from the fcb - doesn't hurt for first */
-  Dmatch_ff.dm_drive = lpFcb->fcb_sftno;
+  if (First)
+  {
+    /* Reconstruct the dirmatch structure from the fcb - doesn't hurt for first */
+    Dmatch_ff.dm_drive = lpFcb->fcb_sftno;
 
-  fmemcpy(Dmatch_ff.dm_name_pat, lpFcb->fcb_fname, FNAME_SIZE + FEXT_SIZE);
-  DosUpFMem((BYTE FAR *) Dmatch_ff.dm_name_pat, FNAME_SIZE + FEXT_SIZE);
+    fmemcpy(Dmatch_ff.dm_name_pat, lpFcb->fcb_fname, FNAME_SIZE + FEXT_SIZE);
+    DosUpFMem((BYTE FAR *) Dmatch_ff.dm_name_pat, FNAME_SIZE + FEXT_SIZE);
 
-  Dmatch_ff.dm_attr_srch = wAttr;
-  Dmatch_ff.dm_entry = lpFcb->fcb_strtclst;
-  Dmatch_ff.dm_dircluster = lpFcb->fcb_dirclst;
+    Dmatch_ff.dm_attr_srch = wAttr;
+    Dmatch_ff.dm_entry = lpFcb->fcb_strtclst;
+    Dmatch_ff.dm_dircluster = lpFcb->fcb_dirclst;
 
-  wAttr = D_ALL;
+    wAttr = D_ALL;
+  }
 
   if ((xfcb FAR *) lpFcb != lpXfcb)
   {


stsp:
`FcbFindFirstNext()` also seems to have a lot
of uses of `Dmatch_ff` instead of `Dmatch_ff_p`.
Could you please experiment with something
more fundamental, for example
```
    dmatch _Dmatch;
    dmatch FAR * Dmatch_p;
    #define Dmatch (*Dmatch_p)
```
and give the same treatment to Dmatch_ff?
And revert 5fb6f0b.

andrewbird:
Okay, I'll write a test for `FcbFindFirstNext()` first to see if it's broken.

andrewbird:
I added tests for FindFirst/FindNext, but although they pass on MS-DOS 6.22, FreeDOS 1.20 is broken with wildcards on fcbfind, so I guess we'll have to fix more than the pointer problem.

andrewbird:
Even before I tweak FCB findfirst/next FDPP crashes with the following
`python3 test/test_dos.py PPDOSGITTestCase.test_fat_fcb_find_simple`

[test.log](https://github.com/stsp/fdpp/files/3962482/test.log)

stsp:
Too many "optimized out", you should enable
extra debug. Plus I think I'd need a test-case,
as I really forgot how this all works, so I don't
think I can find the problem just by the log.
It failed on this:
`  Dmatch_ff.dm_dircluster = lpFcb->fcb_dirclst;`
but, except not using Dmatch_ff_p, I don't
see any problem with that line.

andrewbird:
Here's the simple test case. Create a file called `testa.bat` in the current directory, then run it.

[t2.zip](https://github.com/stsp/fdpp/files/3963804/t2.zip)

andrewbird:
Fixing longhand without the define makes the crash go away. But with the following test, not all the results are found.
[t3.zip](https://github.com/stsp/fdpp/files/3963844/t3.zip)

That's what I see also on FreeDOS 1.20

stsp:
Do you mean this fixes the crash
that define doesn't fix? That would
be very interesting as I suspected
the troubles of parsing (*ptr).memb
expressions.

andrewbird:
Yes the crash went away with the longhand fix, but I may have just screwed something up with the define method. Why find/next should be different to rename/delete where the define method was fine, I don't really know, except that I notice Dmatch_ff is a global instead of local like Dmatch. I'm not a great fan of hiding changes with macros, but I guess it helps keep the diff from FreeDOS small.
I'm working on the missing results now, which seems to be a problem with the pattern wildcard expansion of '*' being converted to '????????' only in findfirst.

andrewbird:
So I have a rough patch that fixes the findnext, but I wonder what the proper fix should be?

```C
diff --git a/kernel/dosfns.c b/kernel/dosfns.c
index 1105142..7b34a4e 100644
--- a/kernel/dosfns.c
+++ b/kernel/dosfns.c
@@ -1020,7 +1020,7 @@ COUNT DosFindNext(void)
 #if 0
   _printf("findnext: %d\n", dmp->dm_drive);
 #endif
-  fmemcpy(&sda_tmp_dm, dmp, 21);
+  //fmemcpy(&sda_tmp_dm, dmp, 21);
 
   /* findnext will always fail on a volume id search or device name */
   if ((sda_tmp_dm.dm_attr_srch & ~(D_RDONLY | D_ARCHIVE | D_DEVICE)) == D_VOLID
diff --git a/kernel/fcbfns.c b/kernel/fcbfns.c
index 956d87f..d46b317 100644
--- a/kernel/fcbfns.c
+++ b/kernel/fcbfns.c
@@ -655,8 +655,11 @@ UBYTE FcbFindFirstNext(xfcb FAR * lpXfcb, BOOL First)
   /* Reconstruct the dirmatch structure from the fcb - doesn't hurt for first */
   Dmatch_ff_p->dm_drive = lpFcb->fcb_sftno;
 
-  fmemcpy(Dmatch_ff_p->dm_name_pat, lpFcb->fcb_fname, FNAME_SIZE + FEXT_SIZE);
-  DosUpFMem((BYTE FAR *) Dmatch_ff_p->dm_name_pat, FNAME_SIZE + FEXT_SIZE);
+  if (First)
+  {
+    fmemcpy(Dmatch_ff_p->dm_name_pat, lpFcb->fcb_fname, FNAME_SIZE + FEXT_SIZE);
+    DosUpFMem((BYTE FAR *) Dmatch_ff_p->dm_name_pat, FNAME_SIZE + FEXT_SIZE);
+  }
 
   Dmatch_ff_p->dm_attr_srch = wAttr;
   Dmatch_ff_p->dm_entry = lpFcb->fcb_strtclst;
```

For instance:
 1/ Holding persistent data in a structure called `sd_tmp_dm` doesn't seem like it was intentional?
 2/ Other fields of `Dmatch_ff` are still changed when called as 'next', and the comment suggests that they shouldn't?
 3/ `fmemcpy()` of 21 bytes, is that still okay in the FDPP regime?

andrewbird:
Mmm, that patch seems to screw things up for FCB rename and delete with wildcards, so definitely wrong.

stsp:
```
C:\test>fcbfind1
Find Operation Success
(TESTA   BAT)
Error reading from drive A: DOS area: general failure
(A)bort, (I)gnore, (R)etry, (F)ail?
```
Drive A: is definitely odd.
But no crash...

andrewbird:
Mmm, I haven't seen that before.

stsp:
Please use this as a work-around for now:
```
diff --git a/kernel/fcbfns.c b/kernel/fcbfns.c
index 82523a9..4852eb5 100644
--- a/kernel/fcbfns.c
+++ b/kernel/fcbfns.c
@@ -643,6 +643,7 @@ UBYTE FcbFindFirstNext(xfcb FAR * lpXfcb, BOOL First)
   /* First, move the dta to a local and change it around to match */
   /* our functions.                                               */
   lpDir = (BYTE FAR *)dta;
+if (First)
   Dmatch_ff_p = MK_FAR(Dmatch_ff);
   dta = Dmatch_ff_p;
 
```

I'll see if something better can be done,
but it should fix the problem too.

stsp:
OK, I know what's going on.
Far emulator reordered cache flushes
with prefetches too optimistically.
If we hint him with a barrier like this:
```
diff --git a/kernel/fcbfns.c b/kernel/fcbfns.c
index 82523a9..4716270 100644
--- a/kernel/fcbfns.c
+++ b/kernel/fcbfns.c
@@ -643,6 +643,7 @@ UBYTE FcbFindFirstNext(xfcb FAR * lpXfcb, BOOL First)
   /* First, move the dta to a local and change it around to match */
   /* our functions.                                               */
   lpDir = (BYTE FAR *)dta;
+Dmatch_ff_p = NULL;
   Dmatch_ff_p = MK_FAR(Dmatch_ff);
   dta = Dmatch_ff_p;
 
```
then it will work properly.
I need to recall how it establishes the
order of operations.

stsp:
I think I fixed this instability.
Please retry your experiments with the
crash, with macros and other things that
behaved weirdly. There was a data corruption
caused by read/write reordering on aliased
pointers. Reordering was happening because
Dmatch_ff_p is static (far emulator haven't
evaluated its life-time and haven't deduced
the right order from it) and aliasing happened
because Dmatch_ff is static, and together that
formed a nasty corner case. Teaching this code
about the static data again (I removed the static
handling long ago) is not productive, so I applied
a hack that detects an alias and adds an extra
read.

andrewbird:
I force pushed this branch again, unfortunately even with your latest fix the defined pointer stuff is still crashing FDPP. See [test.log](https://github.com/stsp/fdpp/files/3964832/test.log)

stsp:
But I used your t3.zip as a test, can
you confirm that at least that works
as expected now?

stsp:
Alsp you were talking about some
crash that happens even w/o your
changes.

andrewbird:
> But I used your t3.zip as a test, 

That's very strange as that's what I tested in the log I just posted
~~~
D:\>c:\fcbfind3                                                                 
                                                                                
                                                                                
fdpp crashed.                                                                   
ERROR: fdpp: abort at ./farptr.hpp:33                                           
Press any key to exit.                                                          
~~~

> Alsp you were talking about some
crash that happens even w/o your
changes.

Yes, so I just did checkout the previous commit 0bec049 and the crash there is in the same place as my HEAD
```
DosUpFMem((BYTE FAR *) Dmatch_ff.dm_name_pat, FNAME_SIZE + FEXT_SIZE);
```
So it seems to me that replacing `Dmatch_ff.dm_name_pat` with the define `(*Dmatch_ff_p).dm_name_pat` is not exactly the same (in terms of FDPP) as `Dmatch_ff_p->dm_name_pat` when I did it by hand.  Is that an accurate assumption?

andrewbird:
With the latest HEAD (with #define in place), just changing the following cures the crash.
```c
diff --git a/kernel/fcbfns.c b/kernel/fcbfns.c
index b9be8b1..4556034 100644
--- a/kernel/fcbfns.c
+++ b/kernel/fcbfns.c
@@ -658,7 +658,7 @@ UBYTE FcbFindFirstNext(xfcb FAR * lpXfcb, BOOL First)
   Dmatch_ff.dm_drive = lpFcb->fcb_sftno;
 
   fmemcpy(Dmatch_ff.dm_name_pat, lpFcb->fcb_fname, FNAME_SIZE + FEXT_SIZE);
-  DosUpFMem((BYTE FAR *) Dmatch_ff.dm_name_pat, FNAME_SIZE + FEXT_SIZE);
+  DosUpFMem((BYTE FAR *) Dmatch_ff_p->dm_name_pat, FNAME_SIZE + FEXT_SIZE);
 
```

stsp:
> Yes, so I just did checkout the previous commit 0bec049 

What is "previous commit 0bec049", a parent of 0bec049,
or itself? Because if you used it itself (rather than its
parent), then it is definitely with your changes.
So could you please confirm that you see the crash
on regular master?

> So it seems to me that replacing Dmatch_ff.dm_name_pat with the define
> (*Dmatch_ff_p).dm_name_pat is not exactly the same (in terms of FDPP) as
> Dmatch_ff_p->dm_name_pat when I did it by hand. Is that an accurate assumption?

Yes, I can think of a different influence of
that on the context below such lines, even
if the lines itself are conceptually similar.
For example I could forget to update the
far cache when parsing `(*Dmatch_ff_p).dm_name_pat`
(because this is a relatively rare form in
freedos, so got little testing), and then there
would be a subtle crash later. Need to check.

> With the latest HEAD (with #define in place), just changing the following cures the crash.

Thanks for the hint, will now check.

andrewbird:
And changing it to this (which I think is what the macro expands to) the crash is back
```c
diff --git a/kernel/fcbfns.c b/kernel/fcbfns.c
index b9be8b1..3517c6b 100644
--- a/kernel/fcbfns.c
+++ b/kernel/fcbfns.c
@@ -658,7 +658,7 @@ UBYTE FcbFindFirstNext(xfcb FAR * lpXfcb, BOOL First)
   Dmatch_ff.dm_drive = lpFcb->fcb_sftno;
 
   fmemcpy(Dmatch_ff.dm_name_pat, lpFcb->fcb_fname, FNAME_SIZE + FEXT_SIZE);
-  DosUpFMem((BYTE FAR *) Dmatch_ff.dm_name_pat, FNAME_SIZE + FEXT_SIZE);
+  DosUpFMem((BYTE FAR *) (*Dmatch_ff_p).dm_name_pat, FNAME_SIZE + FEXT_SIZE);
 
   Dmatch_ff.dm_attr_srch = wAttr;
   Dmatch_ff.dm_entry = lpFcb->fcb_strtclst;
```

andrewbird:
> What is "previous commit 0bec049", a parent of 0bec049,
or itself?

Itself, sorry for not being clear.

> So could you please confirm that you see the crash
on regular master?

Will check later.

stsp:
> Will check later.

I thought this is what you were saying here:
`Even before I tweak FCB findfirst/next FDPP crashes with the following`.
I thought this means "on regular master", or?

andrewbird:
with master it doesn't crash but that's because it still has an unreverted 5fb6f0bca7

I have to go out now, speak later.

stsp:
> with master it doesn't crash

It doesn't now, or didn't yesterday, before my fix?

stsp:
OK, I am starting to understand what you've done. :)
1. You told me (at least it sounded like that) that there
is a crash on master, and provided the test-case. It
wasn't actually on master, but I believed so, reproduced
the crash with your test-case _on master_, and spent
the whole night fixing. Now it appears, by
`Even before I tweak FCB findfirst/next FDPP crashes`
you likely didn't mean "on master", so I had to not test
it on master at all. :(
2. You told (or at least it sounded like) that there is a
crash after you wrapped things with the macro, and
asked if there are any differences. I spent a lot of time
finding them, and was in fact about to commit a fix,
but did a last look on your branch and realized that
it have nothing to do with `Dmatch_ff`. You actually
simply forgot to wrap it, you only wrapped `Dmatch`.
Instead, the crash you've seen was because you did
the revert of 5fb6f0b, not because of macros.

It seems, if someone is fairly convinced that the
bug A can be reproduced with test B, he will eventually
reproduce and fix it no matter whether there was
actually such bug or not. :) That's because, in fact,
it appears the macro produces the different effect,
no matter have you forgot to use it on `Dmatch_ff`
or not. You may be looking for something like this:
```
diff --git a/fdpp/farptr.hpp b/fdpp/farptr.hpp
index 628bb8f..8b92bec 100644
--- a/fdpp/farptr.hpp
+++ b/fdpp/farptr.hpp
@@ -720,6 +720,7 @@ public:
 #define GET_FAR(f) (f).get_far()
 #define GET_PTR(f) (f).get_ptr()
 #define ADJUST_FAR(f) (f).adjust_far()
+#define USE_FP(p) ({ do_store_far(p.get_far()); p; })
 
 #undef NULL
 #define NULL           nullptr
diff --git a/kernel/fcbfns.c b/kernel/fcbfns.c
index 0dc37cb..b5ccf49 100644
--- a/kernel/fcbfns.c
+++ b/kernel/fcbfns.c
@@ -53,8 +53,9 @@ STATIC const char FAR *GetNameField(__FAR(const char) lpFileName,
 #define TestCmnSeps(lpFileName) (*lpFileName && strchr(":;,=+ \t", *lpFileName) != NULL)
 #define TestFieldSeps(lpFileName) ((unsigned char)*lpFileName <= ' ' || strchr("/\\\"[]<>|.:;,=+\t", *lpFileName) != NULL)
 
-static dmatch Dmatch_ff;
+static dmatch _Dmatch_ff;
 static dmatch FAR * Dmatch_ff_p;
+#define Dmatch_ff (*USE_FP(Dmatch_ff_p))
 
 UBYTE FAR *FatGetDrvData(UBYTE drive, UBYTE * pspc, UWORD * bps, UWORD * nc)
 {
@@ -481,7 +482,7 @@ UBYTE FcbDelete(xfcb FAR * lpXfcb)
     int attr = (lpXfcb->xfcb_flag == 0xff ? lpXfcb->xfcb_attrib : D_ALL);
     dmatch _Dmatch;
     dmatch FAR * Dmatch_p;
-#define Dmatch (*Dmatch_p)
+#define Dmatch (*USE_FP(Dmatch_p))
 
     Dmatch_p = MK_FAR(_Dmatch);
     dta = Dmatch_p;
@@ -531,7 +532,7 @@ UBYTE FcbRename(xfcb FAR * lpXfcb)
   {
     dmatch _Dmatch;
     dmatch FAR * Dmatch_p;
-#define Dmatch (*Dmatch_p)
+#define Dmatch (*USE_FP(Dmatch_p))
     COUNT _result;
 
     wAttr = (lpXfcb->xfcb_flag == 0xff ? lpXfcb->xfcb_attrib : D_ALL);
@@ -647,7 +648,7 @@ UBYTE FcbFindFirstNext(xfcb FAR * lpXfcb, BOOL First)
   /* First, move the dta to a local and change it around to match */
   /* our functions.                                               */
   lpDir = (BYTE FAR *)dta;
-  Dmatch_ff_p = MK_FAR(Dmatch_ff);
+  Dmatch_ff_p = MK_FAR(_Dmatch_ff);
   dta = Dmatch_ff_p;
 
   /* Next initialze local variables by moving them from the fcb   */
```

With that hack the things seems stable, no matter
what you say or do, or what I think you say or do. :)

stsp:
I did a major rewrite of that code in a hope
the parsing will now be more predictable.
In particular the macro trick should now work.

andrewbird:
Thanks for the update as using the define macro method now seems to work as expected. I read that you think I forgot to use macros for Dmatch_ff, but I had used them as can be seen in b8af8a3 and before that commit I described how the define method wasn't working for me and had reverted to the longhand method to avoid the crash.
So I rebased my first two commits now (the third you've already applied your own similar one) on your new master, and we have no crashes.
I still can't get the find next to return multiple files, even with your proposed quick fix below
```c
diff --git a/kernel/fcbfns.c b/kernel/fcbfns.c
index dfc1bc6..79fecba 100644
--- a/kernel/fcbfns.c
+++ b/kernel/fcbfns.c
@@ -648,7 +648,8 @@ UBYTE FcbFindFirstNext(xfcb FAR * lpXfcb, BOOL First)
   /* First, move the dta to a local and change it around to match */
   /* our functions.                                               */
   lpDir = (BYTE FAR *)dta;
-  Dmatch_ff_p = MK_FAR(_Dmatch_ff);
+  if (First)
+    Dmatch_ff_p = MK_FAR(_Dmatch_ff);
   dta = Dmatch_ff_p;
 
   /* Next initialze local variables by moving them from the fcb   */
```
I wonder if that's because the if conditionalises the set of `Dmatch_ff_p` and with the revert of 5fb6f0b a (*_Dmatch_ff) is used in its place elsewhere? Is it still required to revert that old fix?

Thinking on a little bit about this I find myself guessing about what exactly you want, and I know in the past you think perhaps I don't try the things you ask, but I do and try to add a little more experimentation too. I think it can get a little frustrating on both sides, so it's probably better if I stick to reporting issues, providing test cases, and you fix them when you have time. I've mentioned before that I do this only for fun, so it's not like I'm waiting on a fix. Does that sound reasonable?

stsp:
> I had used them as can be seen in b8af8a3 and before that commit I described how the define
> method wasn't working for me and had reverted to the longhand method to avoid the crash.

OK but I think the crash was related to the
revert of 5fb6f0b, not because of a macro.
But would be difficult to say for sure now
as the code was rewritten.

> I still can't get the find next to return multiple files, even with your proposed quick fix below

That one should be no longer needed.
If it changes something, its a bug.
I hope the current code won't require the
hacks like that one.
But interestingly, your fcbfind3.com reports
success for me. Can you try on a regular
master first?

> Is it still required to revert that old fix?

This is not strictly required. This was just
a proposal, I hope we can do the trick with
it reverted. Lets isolate the problems and see.

> Thinking on a little bit about this I find myself guessing about what exactly you want

Ah, it's not about what I want, its just that
I wasn't able to properly understand what
you do, on what commits you see the failure
etc, so I tested things not the way you did
and fixed not what you asked for.
I still do not have a clear picture about the
crashes you've seen (which ones were on
master and which were with some changes),
but that's now moot as the code was rewritten.
We just need to start from scratches and
re-document your findings. :)

> I think it can get a little frustrating on both sides

No, its just that the lack of understanding
made me to fix things not on the branches
where you actually have seen the problems.
But from the second attempt I fixed them
on top of your branch, yet now it looks like
again you've seen something else. :) This
is not a problem at all, but the more precise
documenting may help when you test
something not on master. I am not against
any experiments, its just when I am trying
to reproduce them, I always get something
else. :) Perhaps if you see more than one
different crash, it will help to put them into
a different tickets, or come up with some
other instrumentation that will help to re-sync
our environments quickly. And I really hope
this problem will disappear soon, once fdpp
is more stable than it is now. Which definitely
can't be achieved w/o your testing.

andrewbird:
> OK but I think the crash was related to the
revert of 5fb6f0b, not because of a macro.

Yes the crash was caused by the revert, but adding the Dmatch_ff macro should have fixed it again, and it did not.

> But interestingly, your fcbfind3.com reports
success for me.

Run on its own it may look like that, 'Success' is printed on the correct return value of findfirst and its filename printed, but subsequently findnext should print four further file names. The test logic is in the python test that checks for their display. The dos part is the minimum code that allows parsing by python logic.

Here's the current Travis test of FDPP master https://travis-ci.com/andrewbird/dosemu2/builds/141000207 there are no crashes I can see.

So I see two problems remaining now
1/ Fcbdelete needs DMatch macro (FcbRename doesn't because it has been changed to use DMatch_p by the old commit  5fb6f0b)
2/ FcbFindfirstnext doesn't return any results after the first.

stsp:
> Yes the crash was caused by the revert, but adding the Dmatch_ff macro should have fixed it
> again, and it did not.

I suppose there was no reason to revert the macro,
as the crash stayed nevertheless but I was confused
seeing the crash and no macro.
In any case, this is not a problem, the slight problem
was when I tried to fix that on master instead, as
fixing something that isn't there, may take too long. :)

> Run on its own it may look like that,

Could you please provide the reference output that
I should see when things work fine? Also am I right
there are no changes wrt this test with the recent changes?

> 1/ Fcbdelete needs DMatch macro (FcbRename doesn't because it has been
> changed to use DMatch_p by the old commit 5fb6f0b)

But with old commit reverted and the macro in both
places, this problem is fixed, right? If so, that should
be the path.

andrewbird:
I just force pushed  a partial revert of 5fb6f0b that lets the Dmatch macro work for FcbRename in the same way it does for FcbDelete. So only the FcbFindfirstnext to fix now, Here are the tests
~~~
Test PP-DOS-GIT FAT FCB file delete simple                                       ... ok
Test PP-DOS-GIT FAT FCB file delete missing                                      ... ok
Test PP-DOS-GIT FAT FCB file delete wildcard one                                 ... ok
Test PP-DOS-GIT FAT FCB file delete wildcard two                                 ... ok
Test PP-DOS-GIT FAT FCB file delete wildcard three                               ... ok
Test PP-DOS-GIT MFS FCB file delete simple                                       ... ok
Test PP-DOS-GIT MFS FCB file delete missing                                      ... ok
Test PP-DOS-GIT MFS FCB file delete wildcard one                                 ... ok
Test PP-DOS-GIT MFS FCB file delete wildcard two                                 ... ok
Test PP-DOS-GIT MFS FCB file delete wildcard three                               ... ok
Test PP-DOS-GIT FAT FCB file rename simple                                       ... ok
Test PP-DOS-GIT FAT FCB file rename source missing                               ... ok
Test PP-DOS-GIT FAT FCB file rename target exists                                ... ok
Test PP-DOS-GIT FAT FCB file rename wildcard one                                 ... ok
Test PP-DOS-GIT FAT FCB file rename wildcard two                                 ... ok
Test PP-DOS-GIT FAT FCB file rename wildcard three                               ... ok
Test PP-DOS-GIT FAT FCB file rename wildcard four                                ... ok
Test PP-DOS-GIT MFS FCB file rename simple                                       ... ok
Test PP-DOS-GIT MFS FCB file rename source missing                               ... ok
Test PP-DOS-GIT MFS FCB file rename target exists                                ... ok
Test PP-DOS-GIT MFS FCB file rename wildcard one                                 ... ok
Test PP-DOS-GIT MFS FCB file rename wildcard two                                 ... ok
Test PP-DOS-GIT MFS FCB file rename wildcard three                               ... ok
Test PP-DOS-GIT FAT FCB file rename wildcard four                                ... ok
Test PP-DOS-GIT FAT FCB file find simple                                         ... ok
Test PP-DOS-GIT FAT FCB file find missing                                        ... ok
Test PP-DOS-GIT FAT FCB file find wildcard one                                   ... FAIL
Test PP-DOS-GIT FAT FCB file find wildcard two                                   ... FAIL
Test PP-DOS-GIT FAT FCB file find wildcard three                                 ... FAIL
Test PP-DOS-GIT MFS FCB file find simple                                         ... ok
Test PP-DOS-GIT MFS FCB file find missing                                        ... ok
Test PP-DOS-GIT MFS FCB file find wildcard one                                   ... FAIL
Test PP-DOS-GIT MFS FCB file find wildcard two                                   ... FAIL
Test PP-DOS-GIT MFS FCB file find wildcard three                                 ... FAIL
~~~

Here's what you see for 'file find wildcard one' (fcbfind3) on MS-DOS 6.22
~~~
D:\>c:\fcbfind3
Find Operation Success
(THREE   IN )
(ONE     IN )
(FIVE    IN )
(TWO     IN )
(FOUR    IN )

D:\>DIR

 Volume in drive D is IR DXXXXS D
 Directory of D:\

THREE    IN              7 12-16-19   4:48p
ONE      IN              7 12-16-19   4:48p
FIVE     IN              7 12-16-19   4:48p
TWO      IN              7 12-16-19   4:48p
FOUR     IN              7 12-16-19   4:48p
NONE     CTL             7 12-16-19   4:48p
        6 file(s)             42 bytes
                   2,147,450,880 bytes free

D:\>rem end
~~~

stsp:
But I am confused again because earlier you said:
`Thanks for the update as using the define macro method now seems to work as expected.`
And now you say you need only partial revert.
So the macro method still doesn't work as expected
with the full revert? Will it crash?

> Here's what you see for 'file find wildcard one' (fcbfind3) on MS-DOS 6.22

OK, thanks.

andrewbird:
The macro method works exactly as expected now, no crash with the full revert.
Earlier I was looking for reasons why your find next fix below didn't work for me (to return the full set of results), which is why I asked whether we still needed to revert.
```c
diff --git a/kernel/fcbfns.c b/kernel/fcbfns.c
index 82523a9..4852eb5 100644
--- a/kernel/fcbfns.c
+++ b/kernel/fcbfns.c
@@ -643,6 +643,7 @@ UBYTE FcbFindFirstNext(xfcb FAR * lpXfcb, BOOL First)
   /* First, move the dta to a local and change it around to match */
   /* our functions.                                               */
   lpDir = (BYTE FAR *)dta;
+if (First)
   Dmatch_ff_p = MK_FAR(Dmatch_ff);
   dta = Dmatch_ff_p;
 ```
But I still needed some of the revert to make the FcbRename use the macro in the same way as FcbDelete, hence the partial revert.

In any case the partial revert with the above patch didn't fix the lack of further file results, so from my perspective either the full or the partial revert is okay to apply but an additional fix is required to get the full set of file results.

**Comment by stsp** [comment]
> Earlier I was looking for reasons why your find next fix below didn't work for me
> (to return the full set of results), which is why I asked whether we still needed to revert.

Grr, I don't understand. :)
That small "fix" was proposed by me to fix the
crash _on master_. That was because I misunderstood
what you say (you didn't have crash on master). And now -
what that "fix" does for you? In what exactly way it helps?
Does it help in getting all results?

> But I still needed some of the revert to make the FcbRename use the macro in the
> same way as FcbDelete

I don't understand what is "needed some of the revert".
Maybe that means "needed to keep some parts unreverted"?
What exactly is the difference between full and
partial revert wrt visible (to user) functionality?

> to make the FcbRename use the macro in the same way as FcbDelete, hence the partial revert.

Looking into the partial revert, I see you left
the `FcbFindFirstNext()` part unreverted, and
reverted both `FcbDelete` and `FcbRename`.
So whether or not revert is partial or full, there
would be no difference in `FcbDelete` and `FcbRename`.
So why a partial revert, after all? In what way it
is better than the full one, if neither of them crash?

**Comment by andrewbird** [comment]
> That small "fix" was proposed by me to fix the
crash on master.

Oh really, that's not what I understood until now, I thought it was in response to my question
> So I have a rough patch that fixes the findnext, but I wonder what the proper fix should be?

Oh dear how can so many things be misinterpreted!

> And now -
what that "fix" does for you? In what exactly way it helps?
Does it help in getting all results?

(Nothing, It doesn't, No)

> So why a partial revert, after all? In what way it
is better than the full one, if neither of them crash?

Not looking to fix a crash any more, there are none and user functionality is the same either way. My original PR used the longhand method for FcbDelete in the same way 5fb6f0b did for FcbRename. Way back at the start you had given me the quest to revert 5fb6f0b and use the macro method instead, so I was continuing with that. Without the partial revert the macro does nothing in FcbRename.

**Comment by stsp** [comment]
> Oh really, that's not what I understood until now, I thought it was in response to my question

It could! Because with that fix I got success
from your fcbfind3. So I thought its an answer
(but the one that shouldn't be needed any longer),
until you told me the "success" is not a success
at all. :) This is why I was complaining to having
to do a lot of work that, at the end, wasn't what
you asked for.
In any case, it looks strange that you didn't check
whether it actually fixes anything or not, or you would
just drop it.

> (Nothing, It doesn't, No)

I guess you only checked that right now, otherwise
its quite strange that you kept asking about it. :)

> Without the partial revert the macro does nothing in FcbRename.

Yep, but in this case I just find it strange that
you switched to a partial revert from full revert.
If its not any better, then why not to stick with
the full one?

**Comment by stsp** [comment]
OK I agree that something like this should
be done:
```
diff --git a/kernel/fcbfns.c b/kernel/fcbfns.c
index 1778852..ee4d1f0 100644
--- a/kernel/fcbfns.c
+++ b/kernel/fcbfns.c
@@ -649,17 +649,20 @@ UBYTE FcbFindFirstNext(xfcb FAR * lpXfcb, BOOL First)
 
   /* Next initialze local variables by moving them from the fcb   */
   lpFcb = CommonFcbInit(lpXfcb, SecPathName, &FcbDrive);
-  /* Reconstrct the dirmatch structure from the fcb - doesn't hurt for first */
-  Dmatch_ff.dm_drive = lpFcb->fcb_sftno;
+  if (First)
+  {
+    /* Reconstrct the dirmatch structure from the fcb - doesn't hurt for first */
+    Dmatch_ff.dm_drive = lpFcb->fcb_sftno;
 
-  fmemcpy(Dmatch_ff_p->dm_name_pat, lpFcb->fcb_fname, FNAME_SIZE + FEXT_SIZE);
-  DosUpFMem((BYTE FAR *) Dmatch_ff_p->dm_name_pat, FNAME_SIZE + FEXT_SIZE);
+    fmemcpy(Dmatch_ff_p->dm_name_pat, lpFcb->fcb_fname, FNAME_SIZE + FEXT_SIZE);
+    DosUpFMem((BYTE FAR *) Dmatch_ff_p->dm_name_pat, FNAME_SIZE + FEXT_SIZE);
 
-  Dmatch_ff.dm_attr_srch = wAttr;
-  Dmatch_ff.dm_entry = lpFcb->fcb_strtclst;
-  Dmatch_ff.dm_dircluster = lpFcb->fcb_dirclst;
+    Dmatch_ff.dm_attr_srch = wAttr;
+    Dmatch_ff.dm_entry = lpFcb->fcb_strtclst;
+    Dmatch_ff.dm_dircluster = lpFcb->fcb_dirclst;
 
-  wAttr = D_ALL;
+    wAttr = D_ALL;
+  }
 
   if ((xfcb FAR *) lpFcb != lpXfcb)
   {
```

Similar to what you did above.
What regressions does this give?

andrewbird:
> Yep, but in this case I just find it strange that
you switched to a partial revert from full revert.
If its not any better, then why not to stick with
the full one?

This is an example of me second guessing what might be more acceptable to you, i.e. the full revert doesn't help fix the 'missing results' problem, so let's just revert what is necessary for FcbRename to use the macro properly.

andrewbird:
> What regressions does this give?

So let's be explicit, on top of master or what's currently in this PR?

stsp:
But full revert is not a fix.
Its a proposal to keep diffs with freedos smaller
if macros can do the trick.

> So let's be explicit, on top of master or what's currently in this PR?

On top of master.

andrewbird:
Cool that looks better! So I guess you need to decide now whether we do the full revert, or the partial to add to my FCBDelete commit and this patch?

~~~
Test PP-DOS-GIT FAT FCB file delete simple                                       ... FAIL
Test PP-DOS-GIT FAT FCB file delete missing                                      ... ok
Test PP-DOS-GIT FAT FCB file delete wildcard one                                 ... FAIL
Test PP-DOS-GIT FAT FCB file delete wildcard two                                 ... FAIL
Test PP-DOS-GIT FAT FCB file delete wildcard three                               ... FAIL
Test PP-DOS-GIT MFS FCB file delete simple                                       ... FAIL
Test PP-DOS-GIT MFS FCB file delete missing                                      ... ok
Test PP-DOS-GIT MFS FCB file delete wildcard one                                 ... FAIL
Test PP-DOS-GIT MFS FCB file delete wildcard two                                 ... FAIL
Test PP-DOS-GIT MFS FCB file delete wildcard three                               ... FAIL
Test PP-DOS-GIT FAT FCB file rename simple                                       ... ok
Test PP-DOS-GIT FAT FCB file rename source missing                               ... ok
Test PP-DOS-GIT FAT FCB file rename target exists                                ... ok
Test PP-DOS-GIT FAT FCB file rename wildcard one                                 ... ok
Test PP-DOS-GIT FAT FCB file rename wildcard two                                 ... ok
Test PP-DOS-GIT FAT FCB file rename wildcard three                               ... ok
Test PP-DOS-GIT FAT FCB file rename wildcard four                                ... ok
Test PP-DOS-GIT MFS FCB file rename simple                                       ... ok
Test PP-DOS-GIT MFS FCB file rename source missing                               ... ok
Test PP-DOS-GIT MFS FCB file rename target exists                                ... ok
Test PP-DOS-GIT MFS FCB file rename wildcard one                                 ... ok
Test PP-DOS-GIT MFS FCB file rename wildcard two                                 ... ok
Test PP-DOS-GIT MFS FCB file rename wildcard three                               ... ok
Test PP-DOS-GIT FAT FCB file rename wildcard four                                ... ok
Test PP-DOS-GIT FAT FCB file find simple                                         ... ok
Test PP-DOS-GIT FAT FCB file find missing                                        ... ok
Test PP-DOS-GIT FAT FCB file find wildcard one                                   ... ok
Test PP-DOS-GIT FAT FCB file find wildcard two                                   ... ok
Test PP-DOS-GIT FAT FCB file find wildcard three                                 ... ok
Test PP-DOS-GIT MFS FCB file find simple                                         ... ok
Test PP-DOS-GIT MFS FCB file find missing                                        ... ok
Test PP-DOS-GIT MFS FCB file find wildcard one                                   ... ok
Test PP-DOS-GIT MFS FCB file find wildcard two                                   ... ok
Test PP-DOS-GIT MFS FCB file find wildcard three                                 ... ok
~~~

stsp:
I don't see the reason for partial,
that only complicate things. Unless
you know the reason why not, lets
try the full revert.

andrewbird:
Okay I'll test again here first, though!

andrewbird:
Fix checks out on this PR, I needed to do a couple of minor edits on your patch (with attribution) but the tests are good.
~~~
Test PP-DOS-GIT FAT FCB file delete simple                                       ... ok
Test PP-DOS-GIT FAT FCB file delete missing                                      ... ok
Test PP-DOS-GIT FAT FCB file delete wildcard one                                 ... ok
Test PP-DOS-GIT FAT FCB file delete wildcard two                                 ... ok
Test PP-DOS-GIT FAT FCB file delete wildcard three                               ... ok
Test PP-DOS-GIT MFS FCB file delete simple                                       ... ok
Test PP-DOS-GIT MFS FCB file delete missing                                      ... ok
Test PP-DOS-GIT MFS FCB file delete wildcard one                                 ... ok
Test PP-DOS-GIT MFS FCB file delete wildcard two                                 ... ok
Test PP-DOS-GIT MFS FCB file delete wildcard three                               ... ok
Test PP-DOS-GIT FAT FCB file rename simple                                       ... ok
Test PP-DOS-GIT FAT FCB file rename source missing                               ... ok
Test PP-DOS-GIT FAT FCB file rename target exists                                ... ok
Test PP-DOS-GIT FAT FCB file rename wildcard one                                 ... ok
Test PP-DOS-GIT FAT FCB file rename wildcard two                                 ... ok
Test PP-DOS-GIT FAT FCB file rename wildcard three                               ... ok
Test PP-DOS-GIT FAT FCB file rename wildcard four                                ... ok
Test PP-DOS-GIT MFS FCB file rename simple                                       ... ok
Test PP-DOS-GIT MFS FCB file rename source missing                               ... ok
Test PP-DOS-GIT MFS FCB file rename target exists                                ... ok
Test PP-DOS-GIT MFS FCB file rename wildcard one                                 ... ok
Test PP-DOS-GIT MFS FCB file rename wildcard two                                 ... ok
Test PP-DOS-GIT MFS FCB file rename wildcard three                               ... ok
Test PP-DOS-GIT FAT FCB file rename wildcard four                                ... ok
Test PP-DOS-GIT FAT FCB file find simple                                         ... ok
Test PP-DOS-GIT FAT FCB file find missing                                        ... ok
Test PP-DOS-GIT FAT FCB file find wildcard one                                   ... ok
Test PP-DOS-GIT FAT FCB file find wildcard two                                   ... ok
Test PP-DOS-GIT FAT FCB file find wildcard three                                 ... ok
Test PP-DOS-GIT MFS FCB file find simple                                         ... ok
Test PP-DOS-GIT MFS FCB file find missing                                        ... ok
Test PP-DOS-GIT MFS FCB file find wildcard one                                   ... ok
Test PP-DOS-GIT MFS FCB file find wildcard two                                   ... ok
Test PP-DOS-GIT MFS FCB file find wildcard three                                 ... ok
~~~

stsp:
Thanks!
Time to tag the next beta I suppose.

stsp:
Would you like to split that crap into
2 functions, findfirst and next?

andrewbird:
> Would you like to split that crap into
2 functions, findfirst and next?

But that will increase the FreeDOS diff that we just tried so hard to minimise.

stsp:
This function is already almost
completely diffed because of a bug.
Not changing makes sense in the
places which are good by itself but
can't be properly compiled in fdpp.
I.e. changing innocent code is not
good, but changing buggy code is
perfectly fine. The split can just be
viewed as an improved version of
a fix.

andrewbird:
okay, can do, but I have to go out for the rest of the day.

stsp:
The idea of keeping diffs small is actually nearing its EOL.
Initially the purposes were:
1. Allow for an easy resync
2. Make sure we are on a stable basis of freedos,
and only fix fdpp but don't add bugs to freedos,
or we'll never know which one of 2 breaks.

Now resync is not possible as freedos itself
have suddenly changed their sources a lot. :)
And there is no longer a concern about the
stable base, because freedos was found to
be very buggy, and our freedos base is not.
(but fdpp's compiler framework still is, which
is very annoying after 2 years of stabilizing).
I guess as soon as fdpp can be trusted to
compile the reliable code, there would be no
need at all to keep the diffs with freedos small,
and we'll do some big things, like moving
SDA/LoL to UMB. :) As you may guess, this
won't allow the diffs to be small (and honestly,
they are not small already).
Another concern is that writing 64bit DOS in
C++ from scratches (aka dosbox) would
have required much fewer efforts than that
of fdpp. I think in the global picture that could
be a mistake. I.e. it couldn't take 2 years, and
from dosbox devs it definitely didn't took as much.

PR #120, opened 2019-12-17
fcbfns: Split FCBFindFirstNext into two

andrewbird:
Mmm, not sure the split is worth the code duplication? Also should I duplicate the comments and the code in the ifdef 0 block?

### DIFF
--- a/kernel/fcbfns.c
+++ b/kernel/fcbfns.c
@@ -638,7 +638,7 @@ VOID FcbCloseAll()
       DosCloseSft(idx, FALSE);
 }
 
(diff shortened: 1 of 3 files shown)


stsp:
If you care about the duplication, you can do
`StoreFindResults()` or alike to fill in the `lpFcb`.
That will also cover the comment and if0 block
(but nothing else, e.g. I wouldn't put lpDir and
the rest into a common func - too much troubles
for too small duplication).

Other than that, I can apply it also in its current
version. Its still better than the original.

andrewbird:
Apply as is, I think.

stsp:
OK, thanks.

PR #138, opened 2020-05-01
share: Check share table before delete/rename

andrewbird:
Outwardly behaves the same way as MS-DOS 6.22 and so passes the test in https://github.com/dosemu2/dosemu2/pull/1133

Slightly suboptimal in that if deletion is possible it adds to the share table, then immediately removes it, but doing anything else would require a new function to scan the share table without adding to it.

### DIFF
--- a/kernel/share.c
+++ b/kernel/share.c
@@ -183,6 +183,8 @@ static int lock_unlock
 	 unsigned long len,	/* length (in bytes) of region to lock or unlock */
 	 int unlock);		/* non-zero to unlock; zero to lock */
 
+static int is_file_open(char FAR * filename);
+
 	/* Multiplex interrupt handler */
 
 		/* ------------- HOOK ------------- */
@@ -254,6 +256,12 @@ static void interrupt FAR handler2f(iregs FAR *iregs_p)
 				 (   (((unsigned long)iregs.es)<<16) | ((unsigned long)iregs.dx)   ),
 #endif
 				 (iregs.ax & 0x01));
+			return;
+		}
+
+			/* is_file_open (0xa6)*/
+		if ((iregs.ax & 0xff) == 0xa6) {
+			iregs.ax = is_file_open(MK_FP(iregs.ds, iregs.si));
 			return;
 		}
 	}
@@ -585,6 +593,17 @@ static int lock_unlock
 		}
 		return -(0x24);		/* sharing buffer overflow */
 	}
+}
+
+static int is_file_open(char FAR * filename)
+{
+	int i;
+
+	for (i = 0; i < file_table_size; i++) {
+		if (fnmatches(filename, file_table[i].filename))
+			return 1;
+	}
+	return 0;
 }
 
 static void FAR *high_malloc(uint32_t size, int high)
--- a/kernel/dosfns.c
+++ b/kernel/dosfns.c
@@ -1156,10 +1156,13 @@ COUNT DosDelete(const char FAR * path, int attrib)
   if (result & IS_DEVICE)
     return DE_FILENOTFND;
 
+  if (IsShareInstalled(TRUE) && share_is_file_open(PriPathName))
+    return DE_ACCESS;
+
   return dos_delete(PriPathName, attrib);
 }
 
-COUNT DosRenameTrue(const char * path1, const char * path2, int attrib)
+COUNT DosRenameTrue(const char FAR * path1, const char FAR * path2, int attrib)
 {
   if (path1[0] != path2[0])
   {
@@ -1167,6 +1170,9 @@ COUNT DosRenameTrue(const char * path1, const char * path2, int attrib)
   }
   if (FP_OFF(current_ldt) == 0xFFFF || (current_ldt->cdsFlags & CDSNETWDRV))
     return network_redirector(REM_RENAME);
+
+  if (IsShareInstalled(TRUE) && share_is_file_open(path1))
+    return DE_ACCESS;
 
   return dos_rename(path1, path2, attrib);
 }
--- a/kernel/int2f.asm
+++ b/kernel/int2f.asm
@@ -322,6 +322,25 @@ SHARE_LOCK_UNLOCK:
 		mov	ax,0x10a4
 		jmp	short share_common
 
+;           DOS calls this to see if share already has the file marked as open.
+;           Returns:
+;             1 if open
+;             0 if not
+; STATIC WORD share_is_file_open(const char FAR * filename) /* pointer to fully qualified filename */
+		global SHARE_IS_FILE_OPEN
+SHARE_IS_FILE_OPEN:
+		mov	si, ds
+		mov	es, si		; save ds
+		pop	ax		; save return address
+		pop	si		; filename
+		pop	ds		; SEG filename
+		push	ax		; restore return address
+		mov	ax, 0x10a6
+		int	0x2f	     	; returns ax
+		mov	si, es		; restore ds
+		mov	ds, si
+		ret
+
 ; Int 2F Multipurpose Remote System Calls
 ;
 ; added by James Tabor jimtabor@infohwy.com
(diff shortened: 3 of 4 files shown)


stsp:
Maybe just add a special call for that?
Also what about rename?

andrewbird:
Ahh, you've seen it, I just replied in the other thread.

> Maybe just add a special call for that?

I suppose it's a balance of code size and number of times it's likely to be called, against slight inefficiency?

Yes, can add for Rename, I just need to write the test first to see what MS-DOS does.

stsp:
> I suppose it's a balance of code size and number of times it's likely
> to be called, against slight inefficiency?

I don't think its only efficiency.
Atomicity, plus this open_check is quite
convoluted and can even raise the
critical error.
OTOH in a separate call all you need
is to check `fptr->filename && fptr->filename[0]`
to see if its already opened.

andrewbird:
Yes I'm convinced, I'll try to code up such a function.

andrewbird:
I started to add the `share_is_open()` function, but have some questions about parameter passing. In kernel/int2f.asm I copied the code from `share_open_check()` and removed the parameters I didn't need for my new function. I have a couple of questions:
-  why is the return address on the stack and needs to be maintained, whereas other functions like  `share_access_check()` don't have it?
- we maintain DS by saving it in ES, so destroying the value of ES in the process, doesn't this matter?

andrewbird:
Ahh, I see what's happening with `share_access_check()`, it doesn't pop any args off the stack, so the return address is unaffected.

stsp:
> why is the return address on the stack and needs to be maintained, whereas
> other functions like share_access_check() don't have it?

They do, but differently.
You already figured that out though.
Of course you can refactor them to
do the same thing. :)

> we maintain DS by saving it in ES, so destroying the value of ES in the process,
> doesn't this matter?

It may make sense to preserve DS just
to not confuse the DOS compilers (which
we do not support anyway). The preserving
of DS came from commit 5b52510 that
started to alter it. ES usually doesn't matter,
as I think it is used by the DOS compilers
only when accessing the far pointers, so
I just thought clobbering it isn't a big deal.
Anyway, since we do not support the DOS
compilers, this point is moot, and I don't
know if there is any ABI spec for the DOS
compilers that says whether ES is caller
or callee-save. But maybe @bartoldeman 
does. :)

andrewbird:
I pushed a much simplified check function. I've a couple of questions:
- I have the new function checking `if (fptr->filename[0] == '\0')`, not sure whether I really need it with the `fstrncmp()` or not?
- The return of the function is uint16_t which is overkill for a boolean. Ideally it would be uint8_t but since the return value is passed via AX, it becomes messy to stash it and OR in the result in the assembly. Do we care about the oversize?

kernel/share.c
```diff
@@ -587,6 +595,21 @@ static int lock_unlock
...
+static uint16_t is_file_open(char FAR * filename)
+{
+	int i;
+	file_t FAR *fptr;
+
+	for (i = 0; i < file_table_size; i++) {
+		fptr = &file_table[i];
+		if (fptr->filename[0] == '\0')
+			continue;
+		if (fstrncmp(filename, fptr->filename, sizeof(*fptr->filename)) == 0)
```
stsp, line 607: Why asterisk at sizeof?

stsp:
> need it with the fstrncmp() or not?

They use fnmatches() func.

> assembly. Do we care about the oversize?

If you want, you can use eg ZF.
Or you can return in AL (not sure
what exactly became messy).
But no, I don't think we care
about any widening to the machine
word (16 bits in that case).
But don't use stdint types please,
freedos have its own, BYTE, UWORD
etc.

kernel/share.c
andrewbird: I thought `sizeof(fptr->filename)` would give me the size of the pointer, but I now think I'm wrong.

stsp: But with asterisk it would give
sizeof(fptr->filename[0]), which is even worse.
I think sizeof() on far arrays does work in
fdpp right now (but can't be 100% sure
as I reworked it many times, and don't
remember its current state).

bartoldeman:
About calling conventions Agner Fog has a lot of the details (as usual..): https://www.agner.org/optimize/calling_conventions.pdf
Quote: 
"Segment registers: You only have to care about segment registers in 16-bit mode. DS has to be saved and restored if you change it. EScan be changed freely. In DOS programs, ES can have any value. In 16-bit Windows, ES can only have values that are valid segment descriptors. It is not allowed to use ES for other purposes"

ia16-elf-gcc has been a little more creative with DS than most other DOS compilers in that it may change DS mid-function by default (it still assumes DS=SS at function entry and will restore it that way before exiting if needed, unless you specify `-mno-callee-assume-ds-data-segment`)

kernel/share.c
andrewbird: I repushed with `fnmatches()` with which both the pointer and the sizeof are no longer necessary.

stsp:
Thanks for info.
So my "mental feelings" about that
were correct.
Btw, its still not an API spec, so I think
we should take that document as a
reverse-engineered notes, which may
not apply equally to all available compilers
(but likely he tested the most important
ones).

andrewbird:
@bartoldeman thanks very much for that, most helpful.

bartoldeman:
Yes Agner doesn't write specs he just documents his findings, much like Ralf Brown did. Same for all the CPU timings.

He does have links to the ABIs on page 58 of the pdf, but those are for 32-bit and 64-bit, not 16-bit...

andrewbird:
So want to add this fragment to kernel/dosfns.c to do the check on Rename also:
```diff
@@ -1168,6 +1171,9 @@ COUNT DosRenameTrue(const char * path1, const char * path2, int attrib)
   if (FP_OFF(current_ldt) == 0xFFFF || (current_ldt->cdsFlags & CDSNETWDRV))
     return network_redirector(REM_RENAME);
 
+  if (IsShareInstalled(TRUE) && share_is_file_open(path1))
+    return DE_ACCESS;
+
   return dos_rename(path1, path2, attrib);
 }
```

But my new function `share_is_file_open()` takes a FAR *, and `DosRenameTrue()` only has a near const char *. How do I convert?

stsp:
Does it match other DOSes?

andrewbird:
Yes MS-DOS 6.22 refuses to rename an open file.

stsp:
> How do I convert?

Don't convert, just mark the args of DosRenameTrue()
as FAR. Conversions suck.

andrewbird:
Cool, that's what I just guessed!

stsp:
But have you tried as-is?
Near pointer accepts FAR
w/o explicit conversion.

stsp:
Only converting near-->far suck.
Backward is not a problem.

andrewbird:
> But have you tried as-is?

Yes, got some c++ish error about there being no candidate for the function.

stsp:
Please post.

andrewbird:
luckily I found it in my terminal scroll buffer
~~~
[GIT: `git describe`]" -Wall -ggdb3 -O2 -DDEBUG -Wno-format-invalid-specifier -o dosfns.o dosfns.cc
dosfns.cc:1174:33: error: no matching function for call to 'share_is_file_open'
  if (IsShareInstalled(TRUE) && share_is_file_open(path1))
                                ^~~~~~~~~~~~~~~~~~
/clients/common/fdpp.git/kernel/proto.h:496:16: note: candidate function not viable: no known conversion from 'const char *' to 'FarPtr<const char>' for 1st argument
WORD ASMPASCAL share_is_file_open(__FAR(const char) filename);
~~~

stsp:
This is near-->far conversion.
I probably missed what kind of
conversion you were asking about.

andrewbird:
I repushed with the changes that made things work, it passes the tests locally now.

andrewbird:
BTW nice touch to travis, allowing test against different FDPP branch.

stsp:
Thanks!
I guess, time for rc2 soon.

PR #160, opened 2020-12-25
inthndlr: call 2F.1142 for redirector handles passed to 21.7142, call 2F.1142 in SftSeek2 if available  

ecm-pushbx:
Support for the new interface added in dosemu2's MFS https://github.com/dosemu2/dosemu2/pull/1376

### DIFF
--- a/kernel/dosfns.c
+++ b/kernel/dosfns.c
@@ -254,7 +254,57 @@ long DosRWSft(int sft_idx, size_t n, __XFAR(void)bp, int mode)
   return rwblock(sft_idx, bp, n, mode);
 }
 
+static COUNT SftSeek2(int sft_idx, LONG new_pos, unsigned mode, UDWORD * p_result);
 COUNT SftSeek(int sft_idx, LONG new_pos, unsigned mode)
+{
+  UDWORD result;
+  return SftSeek2(sft_idx, new_pos, mode, &result);
+}
+
+static unsigned try_long_compat_remote_seek(sft FAR * s, LONG new_pos,
+	unsigned mode, UDWORD * p_result)
+{
+  struct seek_struct_S {
+    UDWORD member[2];
+  };
+  struct seek_struct_S seek_struct;
+  struct seek_struct_S FAR * ssp;
+  iregs regs = {};
+
+  if (0 == (s->sft_flags & SFT_FSHARED))
+    return 0;
+
+  seek_struct.member[0] = new_pos;
+  if (mode == SEEK_SET)
+    seek_struct.member[1] = 0;
+  else if (0 == (new_pos & 0x80000000))
+    seek_struct.member[1] = 0;
+  else
+    seek_struct.member[1] = 0xFFFFffff;
+
+  regs.es = FP_SEG(s);
+  regs.di = FP_OFF(s);
+  ssp = MK_FAR(seek_struct);
+  regs.ds = FP_SEG_OBJ(&regs, ssp);
+  regs.d.x = FP_OFF_OBJ(&regs, ssp);
+  regs.c.x = mode;
+  regs.a.x = 0x1142;
+  regs.flags |= FLG_CARRY;
+  call_intr(0x2f, MK_FAR_SCP(regs));
+  if (regs.flags & FLG_CARRY)
+    return 0;
+  if (seek_struct.member[1] == 0 || seek_struct.member[1] == 0xFFFFffff)
+  {
+    *p_result = seek_struct.member[0];
+  }
+  else
+  {
+    *p_result = -1;
+  }
+  return 1;
+}
+
+static COUNT SftSeek2(int sft_idx, LONG new_pos, unsigned mode, UDWORD * p_result)
 {
   sft FAR *s = idx_to_sft(sft_idx);
   if (FP_OFF(s) == (UWORD) -1)
@@ -270,6 +320,10 @@ COUNT SftSeek(int sft_idx, LONG new_pos, unsigned mode)
   if (s->sft_flags & SFT_FDEVICE)
   {
     new_pos = 0;
+  }
+  else if (try_long_compat_remote_seek(s, new_pos, mode, p_result))
+  {
+    return SUCCESS;
   }
   else if (mode == SEEK_CUR)
   {
@@ -291,17 +345,19 @@ COUNT SftSeek(int sft_idx, LONG new_pos, unsigned mode)
   }
 
   s->sft_posit = new_pos;
+  *p_result = new_pos;
   return SUCCESS;
 }
 
 ULONG DosSeek(unsigned hndl, LONG new_pos, COUNT mode, COUNT *rc)
 {
   int sft_idx = get_sft_idx(hndl);
+  UDWORD result;
 
   /* Get the SFT block that contains the SFT      */
-  *rc = SftSeek(sft_idx, new_pos, mode);
+  *rc = SftSeek2(sft_idx, new_pos, mode, &result);
   if (*rc == SUCCESS)
-    return idx_to_sft(sft_idx)->sft_posit;
+    return result;
   return *rc;
 }
 
--- a/kernel/inthndlr.c
+++ b/kernel/inthndlr.c
@@ -1555,10 +1555,46 @@ dispatch:
 
       /* case 0x6d and above not implemented : see default; return AL=0 */
 
-    /* LFN API, support only 0xa6 for now */
+    /* LFN API, support only 0x42 and 0xA6 for now */
     case 0x71:
       switch (lr.AL)
       {
+        case 0x42: {
+          iregs saved_r;
+          sft FAR *s;
+          unsigned char idx;
+
+          if (r->BX >= psp->ps_maxfiles)
+          {
+            rc = DE_INVLDHNDL;
+            goto error_exit;
+          }
+          idx = psp->ps_filetab[r->BX];
+          s = idx_to_sft(idx);
+          if (s == (sft FAR *)-1)
+          {
+            rc = DE_INVLDHNDL;
+            goto error_exit;
+          }
+          if (!(s->sft_flags & SFT_FSHARED))
+            goto unsupp;    /* unsupported on local fs yet */
+          /* call to redirector */
+          saved_r = *r;
+          r->ES = FP_SEG(s);
+          r->DI = FP_OFF(s);
+          r->flags |= FLG_CARRY;
+          r->AX = 0x1142;
+          call_intr(0x2f, r);
+          if (!(r->flags & FLG_CARRY)) {
+            r->ES = saved_r.ES;
+            r->DI = saved_r.DI;
+            goto real_exit;
+          }
+          /* carry still set - unhandled */
+          *r = saved_r;
+          goto unsupp;
+          break;
+        }
         case 0xa6: {
           iregs saved_r;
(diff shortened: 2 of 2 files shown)


ecm-pushbx:
Builds fine but untested yet, due to my problems getting to run fdpp.

ecm-pushbx:
Tested with `dosemu -I "lfn_support off"` now, works.

kernel/inthndlr.c
```diff
@@ -1555,10 +1555,46 @@ VOID ASMCFUNC int21_service(iregs FAR * r)
 
       /* case 0x6d and above not implemented : see default; return AL=0 */
 
-    /* LFN API, support only 0xa6 for now */
+    /* LFN API, support only 0x42 and 0xA6 for now */
```
stsp, line 1558: Except that 0x42 is fat32 not lfn...
But whatever.

stsp:
This misses the SEEK_CUR hook though.
And overall I am quite unhappy with it being
71 instead of 73, but I guess that's the life. :)

ecm-pushbx:
I'm trying to write a wrapper around 1142h that goes into SftSeek. How do I allocate an array (two 32-bit integers) that I can take a far pointer off and pass to the redirector interface?

stsp:
For what??
Can you just adjust SftSeek to call
1142 when needed?

ecm-pushbx:
That's what I am trying. Here's my attempt so far, though I expect it doesn't compile yet: https://ulukai.org/ecm/20201225.3

The seek_struct must be allocated to DOS memory so I can initialise it, take a far pointer, pass it to 1142h, then read the result that 1142h wrote to the structure.

If 1142h is not supported for this handle, or the handle is local or a device, then SftSeek2 falls back to its original handling.

stsp:
There are 2 ways of doing that.
Have a helper function in asm that
will take the far pointer and pass it
to 1142.
Exploit the fdpp-specific tricks, see
LBA_Get_Drive_Parameters()
as an example:
```
  bp = MK_FAR(lba_bios_parameters);
  regs.si = FP_OFF_OBJ(&regs, bp);
  regs.ds = FP_SEG_OBJ(&regs, bp);
```

IIRC you need to have bp and lba_bios_parameters
to have the similar life range:
```
  struct _bios_LBA_disk_parameterS lba_bios_parameters;
  struct _bios_LBA_disk_parameterS FAR *bp;
```

stsp:
regs.si = FP_OFF_OBJ(&regs, bp);
is a way to tell fdpp code flow analyzer
that bp and regs are dependent, and
so accesses to "regs" may clobber "*bp".

ecm-pushbx:
Finished it. Note to self: fdpp's "unsigned long" is a 64-bit type. I ended up using uint32_t in the structure definition to fix this.

kernel/dosfns.c
```diff
@@ -271,6 +319,10 @@ COUNT SftSeek(int sft_idx, LONG new_pos, unsigned mode)
   {
     new_pos = 0;
   }
+  else if (try_long_compat_remote_seek(s, new_pos, mode, p_result))
+  {
+    return SUCCESS;
+  }
   else if (mode == SEEK_CUR)
   {
```
stsp, line 327: Any reason you haven't made it
specific to SEEK_CUR?

kernel/dosfns.c
```diff
@@ -255,6 +255,54 @@ long DosRWSft(int sft_idx, size_t n, __XFAR(void)bp, int mode)
...
+    *p_result = seek_struct.member[0];
+  }
+  else
+  {
+    *p_result = -1;
+  }
+  return 1;
+}
+
+COUNT SftSeek2(int sft_idx, LONG new_pos, unsigned mode, unsigned long * p_result)
```
stsp, line 305: Please use freedos macros like
UWORD, UDWORD.

kernel/dosfns.c
ecm-pushbx: I think it is cleaner this way. And we don't depend on the redirector support for the 32-bit seek from EOF service. (I'm not certain whether that is correct in mfs now.) Also, for some corner cases it is better to always call into 1142h. Eg, what if the 32-bit seek ends up with sft_position = FFFFffffh?

kernel/proto.h
```diff
@@ -83,6 +83,7 @@ BOOL check_break(void);
 UCOUNT GenericReadSft(__FAR(sft) sftp, UCOUNT n,__FAR(void) bp,
                       COUNT * err, BOOL force_binary);
 COUNT SftSeek(int sft_idx, LONG new_pos, unsigned mode);
+COUNT SftSeek2(int sft_idx, LONG new_pos, unsigned mode, unsigned long * p_result);
```
stsp, line 86: Why?
Shouldn't it be static?

kernel/dosfns.c
```diff
@@ -255,6 +255,54 @@ long DosRWSft(int sft_idx, size_t n, __XFAR(void)bp, int mode)
...
+  struct seek_struct_S FAR * ssp;
+  seek_struct.member[0] = new_pos;
+  if (mode == SEEK_SET)
+    seek_struct.member[1] = 0;
+  else if (0 == (new_pos & 0x80000000))
+    seek_struct.member[1] = 0;
+  else
+    seek_struct.member[1] = 0xFFFFffff;
+
+  iregs regs = {};
```
stsp, line 282: Would be better to move that up to
the declarations.

kernel/dosfns.c
ecm-pushbx: Done.

kernel/dosfns.c
```diff
@@ -255,6 +255,54 @@ long DosRWSft(int sft_idx, size_t n, __XFAR(void)bp, int mode)
...
+}
+
+static unsigned try_long_compat_remote_seek(sft FAR * s, LONG new_pos,
+	unsigned mode, unsigned long * p_result)
+{
+  if (0 == (s->sft_flags & SFT_FSHARED))
+    return 0;
+
+  struct seek_struct_S {
+    uint32_t member[2];
```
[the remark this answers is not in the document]
ecm-pushbx: Done.

kernel/proto.h
ecm-pushbx: I didn't detect that DosSeek is in the same file. I'm about to amend the commit to make this static, yes.

kernel/dosfns.c
ecm-pushbx: Done.

kernel/proto.h
ecm-pushbx: Done.

ecm-pushbx:
Something is odd here. My updates don't show up correctly.

ecm-pushbx:
Got it, forgot the `-a` switch to `git commit --amend`

stsp:
Thanks!
Please consider adding a test-case
for that functionality.

andrewbird:
I was thinking about adding a test case for this on dosemu, but I'm not really clear about the calling convention and where it's expected to work (just MFS or FAT too?). I see the new function is int21/7142 but beyond that what input should I give it and what output should I expect? Does it work on small seeks too?

stsp:
This is redirector-only thing.
You pass "file handle" in BX and
a pointer to 64bit offset in DS:DX.
CX is SEEK_SET/CUR/END.

andrewbird:
Cool, thanks. And where do I see the resultant position returned?

stsp:
At the same place where the original was.

andrewbird:
Thanks

andrewbird:
So is there anything I need to do to 'enable' the new 7142 command? In my test at https://github.com/andrewbird/dosemu2/commit/b97529ffac22e7c5a15dd9a7da8a47a5b3e17b92 I see that the new redirector call is used on normal djgpp lseeks

~~~
MFS: Entering dos_fs_redirect, FN=42, 'Long seek (extension)'                   
MFS: selecting drive                                                            
MFS: selected drive 23: /tmp/                                                   
MFS: long seek                                                                  
found /tmp/lfsfiles.tst on fd 37                                                
cl=00h seek=0000000Ah sftposition=00000000h                                     
fseek=00000000h fsize=180000000h                                                
result seek=0000000Ah                                                           
MFS: Finished dos_fs_redirect                                                   
~~~
but when I call the new 7142 command I get CF set and AX=7100 not implemented, and the call is never seen by the redirector.

Using FDPP with 
`python3 test/test_dos.py PPDOSGITTestCase.test_lfs_file_seek_tell_mfs_end`

stsp:
I added a few blind/untested patches
to make it supported. As it was indeed
unsupported.

andrewbird:
Thanks, I'll give it a try later today (just new year here!) Hope it will be a better one for all!

andrewbird:
So that's a lot better, I'm now seeing the call to the redirector (that actually surprises me as I thought only SEEK_END would get passed to the redirector), but I'm seeing two calls. The final posn seems to be correct on the first, but something odd on the second.
~~~
MFS: Entering dos_fs_redirect, FN=42, 'Long seek (extension)'                   
MFS: selecting drive                                                            
MFS: selected drive 23: /tmp/                                                   
MFS: long seek                                                                  
found /tmp/lfsfiles.tst on fd 37                                                
cl=00h seek=0000000Ah sftposition=00000000h                                     
fseek=00000000h fsize=180000000h                                                
result seek=0000000Ah                                                           
MFS: Finished dos_fs_redirect                                                   
MFS: Entering dos_fs_redirect, FN=42, 'Long seek (extension)'                   
MFS: selecting drive                                                            
MFS: selected drive 23: /tmp/                                                   
MFS: long seek                                                                  
found /tmp/lfsfiles.tst on fd 37                                                
cl=00h seek=154B7B850E875D3h sftposition=0000000Ah                              
fseek=0000000Ah fsize=180000000h                                                
result seek=154B7B850E875D3h                                                    
MFS: Finished dos_fs_redirect                                                   
~~~

And the test that generated it
`python3 test/test_dos.py PPDOSGITTestCase.test_lfs_file_seek_tell_mfs_set`

stsp:
You see 2 long seeks likely because
normal seek also goes there.

andrewbird:
okay I'll investigate tomorrow why the return value is not what I expect, thanks!

stsp:
Should now work.

andrewbird:
Yes, those latest changes are good. Now works correctly with SEEKSET, SEEKCUR and SEEKEND correctly on FDPP, MSDOS 6.22 and DRDOS 7.01.

stsp:
So your tests seems to be the only thing
that keeps our layered approach alive. :)
Without tests, every new addition misses
some of 10+ existing layers where it have
to be supported.
Maybe the initial ECMs patch, where the
whole thing was implemented in a single
place, is indeed easier to comprehend.

ecm-pushbx:
> This is redirector-only thing.
> You pass "file handle" in BX and
> a pointer to 64bit offset in DS:DX.
> CX is SEEK_SET/CUR/END.

Sorry, I found this thread's updates only now.

The seek origin is actually (as I also commented on one of Andrew's commits) in CL, not CX.

And it is mostly redirector-only, the only other support currently known is Enhanced DR-DOS on a FAT+ FS partition. That originated the 7142h call.

andrewbird:
> The seek origin is actually (as I also commented on one of Andrew's commits) in CL, not CX.

I have an update for my original comment in light of your clarification in an as yet unsubmitted PR

Thanks

PR #167, opened 2021-02-23
Test: Setup Github actions

andrewbird:
1/ Make build environment more generic
  a) Rename scripts
  b) Resurrect package install script
  c) Flip Travis build to use focal and use corresponding packages
  d) Set git user info for quiet reflog of temporary tag

2/ Add github workflow

See https://github.com/andrewbird/fdpp/runs/1963484584?check_suite_focus=true for example run

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -1,0 +1,25 @@
+name: Build
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+      - ready_for_review
+      - reopened
+      - synchronize
+  push:
+
+jobs:
+  build:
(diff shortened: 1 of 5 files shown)


stsp:
Should this be applied?

andrewbird:
Yes, Github Actions seems fine for this.

I've had no human reply from Travis support regarding the request for Open Source Credits after 5 days, so I think I should work on getting Dosemu2 building/testing on there too.

stsp:
Wow, seems working!
Lets see. :)

stsp:
Is it now unlimited?

andrewbird:
> Is it now unlimited?

2000 Action minutes per month (I've done a rough calc that we use about 1500) see https://github.com/pricing

PR #171, opened 2021-03-10
Make: Avoid calling git describe repeatedly

andrewbird:
We don't expect it to change during compilation so let's only call it
once.


stsp:
Thanks.

PR #175, opened 2021-05-01
Test: Avoid building on Travis

andrewbird:
Since we don't actually test anything with CI on FDPP, it's pointless
wasting Travis credits just to check the build as we can do that with
Github Actions. Disable the Travis test by only triggering on cron
events (there are none configured).


stsp:
Thanks.

PR #182, opened 2021-09-08
make: Avoid building thunk_gen with C++

andrewbird:
Removes the following warnings:
  lex /clients/common/fdpp.git/fdpp/parsers/thunk_gen.l
  bison -d /clients/common/fdpp.git/fdpp/parsers/thunk_gen.y
  clang++ -g -o thunk_gen lex.yy.c thunk_gen.tab.c
  clang: warning: treating 'c' input as 'c++' when in C++ mode, this
  behavior is deprecated [-Wdeprecated]
  clang: warning: treating 'c' input as 'c++' when in C++ mode, this
  behavior is deprecated [-Wdeprecated]

### DIFF
--- a/fdpp/parsers/makefile
+++ b/fdpp/parsers/makefile
@@ -4,7 +4,7 @@ srcdir ?= $(CURDIR)
 include $(srcdir)/../clang.mak
 
 thunk_gen: lex.yy.c thunk_gen.tab.c
-	$(CLC) -g -o $@ $^
+	clang -g -o $@ $^
 
 thunk_gen.tab.c: $(srcdir)/thunk_gen.y
 	bison -d $^


stsp:
I committed a slightly
different fix, as there was
a missing build rule that
caused this to happen.
But I can't reproduce that
anyway because for me
`make` defines CXX variable
internally, and so the parser
gets built with g++ instead
of clang++...
Not sure what can be the
good way of fixing that.

andrewbird:
Strangely I found that making within parsers directory used g++ but going to repo root it used clang++. Perhaps that helps?

stsp:
Yes, that helps, thanks.
Should now be fixed.

Issue #200, opened 2022-07-15
FCB (int 21h func 29h): should keep parsing name even if drive letter invalid


tkchia:
On both MS-DOS 2.11 and MS-DOS 5.0, the "parse filename to FCB" syscall (`int 0x21`, function `0x29`), when it encounters an invalid drive letter, will still continue parsing the file name, and return with `al` = `0xff` _and_ `ds:si` pointing after the parsed file name.

E.g. given an input like `z:foo.bar,baz` — assuming `z` is not a valid drive letter — the syscall will return `al` = `0xff` and `ds:si` pointing at the comma.

FreeDOS++ should probably do likewise.  I have already submitted a proposed patch to the upstream FreeDOS kernel project, at https://github.com/FDOS/kernel/pull/82 .

(**Edit:** I also submitted another patch upstream at https://github.com/FDOS/kernel/pull/83 which fixes a problem in my earlier patch.)

Thank you!

stsp:
Andrew, am I remember correctly
that we already did that?

andrewbird:
Your memory must be better than mine, I have no recollection of this topic. Perhaps I should write a test to see what the other DOSes do.

stsp:
But I suppose @tkchia already has such
test?

tkchia:
Hello @stsp, hello @andrewbird,

The test code for this particular behaviour for `int 0x21`, `ah` = `0x29`, is now in my `libi86` master branch &mdash; though the test is disabled for now.  Let me know if you would like me to enable the test.  Thank you!

stsp:
Ping.
Andrew, have you looked at a test?

stsp:
@tkchia maybe you can just
check fdpp yourself and submit
a patch if needed?

tkchia:
Hello @stsp,

> @tkchia maybe you can just check fdpp yourself and submit a patch if needed?

Hmm... let me look into doing that.  Thank you!

PR #201, opened 2022-07-21
FCB (int 21h func 29h): should keep parsing name even if drive letter invalid

tkchia:
Should address https://github.com/dosemu2/fdpp/issues/200 .

### DIFF
--- a/kernel/fcbfns.c
+++ b/kernel/fcbfns.c
@@ -89,7 +89,7 @@ UBYTE FAR *FatGetDrvData(UBYTE drive, UBYTE * pspc, UWORD * bps, UWORD * nc)
 #ifndef IPL
 UWORD FcbParseFname(UBYTE *wTestMode, const char FAR * lpFileName, fcb FAR * lpFcb)
 {
-  WORD wRetCodeName = FALSE, wRetCodeExt = FALSE;
+  WORD wRetCodeDrive = FALSE, wRetCodeName = FALSE, wRetCodeExt = FALSE;
 
   /* pjv -- ExtFcbToFcb?                                          */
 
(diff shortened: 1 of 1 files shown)


tkchia:
I had submitted similar patches to the upstream FreeDOS kernel, as https://github.com/FDOS/kernel/pull/82 and https://github.com/FDOS/kernel/pull/83 .  I had also tested them briefly under QEMU.

I also tried to test this particular patch under `fdpp` specifically, but I am not sure how to properly do so.  If I run something like
```
$ FDPP_KERNEL_DIR="`pwd`/fdpp" dosemu -K . -tde
```
after building the new `fdpp`, then `dosemu2` hangs:
```
ERROR: libao: unable to open output device
dosemu2 2.0pre9 Configured: 2022-02-21 02:17:09 +0000
Get the latest code at http://dosemu2.github.io/dosemu2
Submit Bugs via https://github.com/dosemu2/dosemu2/issues
Ask for help in mail list: linux-msdos@vger.kernel.org
This program comes with ABSOLUTELY NO WARRANTY.
This is free software, GPL v2 (or any later version) distribution conditions.
```

Thank you!

stsp:
Seems to pass the test so
applied.
Thanks!

stsp:
> after building the new fdpp, then dosemu2 hangs:

This usually means that you have
libfdpp.so elsewhere (in /usr/*) and
it was built with different fdpp.
In that case you may try to also set
LD_LIBRARY_PATH="pwd/fdpp"
The problem is missing SEGRELATIVE
relocs in nasm. The relocation table
is hardcoded in libfdpp.so so if you
have it from another build, it will corrupt
the elf file.
We really need SEGRELATIVE support.

PR #214, opened 2023-03-02
Update kernel.ld

c-logic:
change POS of gnu build id to stop
debian 11 overlapping warnung

### DIFF
--- a/fdpp/kernel.ld
+++ b/fdpp/kernel.ld
@@ -14,7 +14,11 @@ SECTIONS
     .ptext 0 : {
 		*(PSP)
     }
-
+    /* for build-id */
+    .note.gnu.build-id  : {
+		*(.note.gnu.build-id)
+	}
+    /* for solib */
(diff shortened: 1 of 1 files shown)


fdpp/kernel.ld
```diff
@@ -14,7 +14,11 @@ SECTIONS
     .ptext 0 : {
 		*(PSP)
     }
-
+    /* for build-id */
+    .note.gnu.build-id  : {
+		*(.note.gnu.build-id)
+	}
+    /* for solib */
```
andrewbird, line 21: Did you mean the last comment to move too?

andrewbird:
I don't have Debian 11, and was unable to get Ubuntu Jammy to emit the error, so I'm unable to test. Would `-Wl,--build-id=none` help instead?

stsp:
Hmm, good point!
Removed build-id, we only need it for libfdpp.so.
Lets see if it helps.

PR #232, opened 2023-10-29
Rel2

stsp:
segelf relocations support.


stsp:
I disabled the arm64 build
on LP for now. fdpp should
be ported to arm once again. :(

PR #234, opened 2023-11-06
CI: test using test suite and dosemu from git

andrewbird:
I had a few moments spare so hopefully this is helpful.


stsp:
Thanks!
Although I really don't remember
if fdpp have ever regressed some
particular test.

PR #242, opened 2023-12-08
Ci 58

andrewbird:


configure.meson
```diff
@@ -1,12 +1,11 @@
-#!/bin/sh
+#!/bin/bash
```
stsp, line 1: Why?

andrewbird: because echo `-e` is used

configure.meson
```diff
@@ -1,12 +1,11 @@
...
 
 # set -xv
 
-if [ -f meson.build ]; then
+if [ ! -d build ] ; then
```
stsp, line 5: Mm, in this case I think you need
to use `$1` here. I.e. do not hard-code
the dir name.

configure.meson
stsp:
```
$ which echo
/usr/bin/echo
$ /usr/bin/echo -e "\taa"
	aa
```

So are you sure?

configure.meson
stsp: Or actually just create the build dir
right here?

configure.meson
andrewbird: The shell (sh) built-in, which gets used in preference to any external command, doesn't support it. The reason it worked for you is that there was no `#!` at the top of the script, so the script was run as a child of your current shell (presumably bash or some other enhanced shell). e.g.
~~~
ajb@calypso:/clients/common/fdpp.git$ sh
$ echo -e "\tbob"
-e 	bob
$ 
~~~

configure.meson
andrewbird: In fact `meson setup` will create it if necessary.

configure.meson
andrewbird: Dropped the formatting and bash requirement.

configure.meson
andrewbird: I modded it now  to just see that the arg is not empty.

stsp:
Thanks!

PR #246, opened 2024-02-18
CI: Update to use current builtin actions

andrewbird:
Older actions use deprecated node16, so update to avoid the warning.


stsp:
Thanks!

PR #249, opened 2024-03-09
A0 01

andrewbird:
For discussion


kernel/network.c
```diff
@@ -156,17 +156,15 @@ BYTE remote_getfree_11a3(void FAR *cds, void *dst)
...
     iregs regs = {};
-    struct xgetvolumeinfo s;
     struct xgetvolumeinfo FAR *p;
 
-    p = MK_FAR(s);
+    p = MK_FAR(*s);
 
     regs.a.x = 0x11a0;
-    regs.c.x = sizeof(s);
+    regs.c.x = sizeof(*s);
```
stsp, line 167: So if that doesn't compile, you
can pass `s_ptr` and do
`*s_ptr = s;` at the end.

andrewbird: It's compiling okay now, thanks.

stsp: Good that I omitted porting this
horrible templated compiler to
dj64dev. :) Fixing the oddities like
this one, could be a nightmare.

andrewbird:
What are your thoughts about the `$_lfn_support = (on)` setting in dosemu.conf? I can see it made sense when lfn support was processed before any TSR, but now dosemu's lfn is only a fallback is it still relevant?

stsp:
What do you propose?

andrewbird:
I was thinking of removal of that config option, but I suppose it may be too early. I'll just need to disable the test for it under FDPP.

stsp:
Yes, I think its mainly historic.
There were times when some DOSes
were unbootable once you enable
lfn support (eg 198bc9c), and even
in dosemu2 history we have the wtf
things like f1002e0 13f4867 bf281b9

But I guess disabling LFNs should
never be a cure, esp since we do
support LFNs in comcom64.
It can also be moved to emuconf.com
to allow disabling dynamically (in
case some program fails with LFNs).
So the question is how "deeply" to
remove it. Removing from dosemu.conf
sounds good, maybe also from parsers,
but leave in the code and eventually
add to emuconf.com?
@bartoldeman do you have an opinion
on that?

stsp:
Ah, commit hashes didn't work,
as we are in fdpp, not dosemu2...

andrewbird:
That would be
https://github.com/dosemu2/dosemu2/commit/198bc9c
https://github.com/dosemu2/dosemu2/commit/f1002e0
https://github.com/dosemu2/dosemu2/commit/13f4867
https://github.com/dosemu2/dosemu2/commit/bf281b9

PR #252, opened 2024-10-12
bpb: Convert sz string to 83 for volume name

andrewbird:
It passed the tests before only because I used a test string of 8 characters, whereas a longer one would contain the dot, and shorter padding spaces.


stsp:
Thanks!
Would be good to update also the test then.

andrewbird:
Oh yes, I've done it locally but not pushed yet.

PR #281, opened 2025-08-24
Cds

stsp:
See https://github.com/dosemu2/dosemu2/issues/838

### DIFF
--- a/kernel/main.c
+++ b/kernel/main.c
@@ -55,7 +55,8 @@ STATIC VOID update_dcb(struct dhdr FAR *);
 STATIC VOID init_kernel(VOID);
 STATIC VOID signon(VOID);
 STATIC VOID kernel(VOID);
-STATIC VOID FsConfig(BOOL reinit);
+STATIC VOID FsConfig(VOID);
+STATIC VOID InitCds(VOID);
 STATIC VOID InitPrinters(VOID);
 STATIC VOID InitSerialPorts(VOID);
 #ifndef FDPP
@@ -405,7 +406,8 @@ STATIC void init_kernel(void)
     update_dcb(&blk_dev);
 
   /* Now config the temporary file system */
-  FsConfig(0);
+  InitCds();
+  FsConfig();
 
   /* Now process CONFIG.SYS     */
   DoConfig(0);
@@ -425,7 +427,8 @@ STATIC void init_kernel(void)
   PostConfig();
 
   /* Init the file system one more time     */
-  FsConfig(1);
+  fmemcpy(LoL->_CDSp, old_CDSp, LoL->_lastdrive * sizeof(struct cds));
+  FsConfig();
 
   configDone();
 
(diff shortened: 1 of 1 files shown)


ecm-pushbx:
I think update_dcb should add the drive to the CDS, so the user can load subsequent drivers or executables from the new block device drive.

I don't understand what you're doing with this mask stuff.

ecm-pushbx:
Also:

> Network drives may not have DPB, although in most cases they do.

I think in most cases redirected drives actually don't have DPBs. dosemu2 mfs with the boot drives is an outlier.

ecm-pushbx:
I added the old fdpp behaviour to https://hg.pushbx.org/ecm/msdos4/shortlog/54cb7afd47e5

Ie, full CDS is only initialised once, then every block device adds only its own drives with older drives preserved, and the final relocation also preserves all drives.

Not done yet: Skip redirector drives during block device setup, and allow to make the initial CDS larger than needed to allow for redirectors to grab new drives.

INSTALLFIRST= is like DR-DOS's INSTALL=, INSTALLMID= is like lDOS default INSTALL=, COMPAT=INSTALLASLAST or COMPAT=INSTALLASFIRST modifies how lDOS handles INSTALL= (as INSTALLFIRST=, INSTALLMID=, or INSTALLLAST=).

stsp:
> What exactly do you mean by "that"?

Adding drive to CDS array.

> I do think I'll add something for early CDS size

What can that be?
Lastdrive=26?
That's cheating. :)

ecm-pushbx:
> > What exactly do you mean by "that"?
> 
> Adding drive to CDS array.

Yes, I do want that in update_dcb.

> > I do think I'll add something for early CDS size
> 
> What can that be? Lastdrive=26? That's cheating. :)

I was thinking more like EARLYLASTDRIVE=Z:, then the final LASTDRIVE only takes effect if it is smaller than the last in-use drive.

stsp:
Thats precisely what freedos does.
But this is not how lastdrive should
grow. This way you only shrink.

Воскресенье, 24 августа 2025 г получено от E. C. Masloch:

ecm-pushbx:
Nice. Nitpick: The DPB end of chain pointer should check just the offset for -1 (ie, 0FFFFh), not the entire 32-bit pointer. Probably fine if everyone sets the entire pointer in your kernel, but there is such a thing as DEVLOAD so may be better to fix this.

stsp:
Why not to fix this in a
freedos then?
Done.

stsp:
So can you click Approve?
Otherwise github awards
me with "you wanted - you
get it" badge for ignoring
reviewers.

ecm-pushbx:
> Why not to fix this in a freedos then? Done.

It should be done, yes. But I have other things to do right now.



> So can you click Approve? Otherwise github awards me with "you wanted - you get it" badge for ignoring reviewers.

Oh, I'll try it. Not sure how this works.

ecm-pushbx (commented):
Looks good.

ecm-pushbx:
Did I do it right?

stsp:
Perfectly. :)

PR #299, opened 2026-07-05
Debug msgs

andrewbird:
1/ Some changes to make the debugging to dosemu logging compile after enabling lots of DEBUG_* defines.
2/ Tidy up some of the messages now that the logging function prefixes every write.
3/ Tweak the CI to collect and archive any failure logs that may occur.

### DIFF
--- a/fdpp/thunks.cc
+++ b/fdpp/thunks.cc
@@ -233,9 +233,9 @@ void fdvprintf(const char *format, va_list vl)
     fdpp->print(FDPP_PRINT_TERMINAL, format, vl);
 }
 
-static void fdlogvprintf(const char *format, va_list vl)
+static void fdlogvprintf(int prio, const char *format, va_list vl)
 {
-    fdpp->print(FDPP_PRINT_LOG, format, vl);
+    fdpp->print(prio, format, vl);
 }
 
 void fdprintf(const char *format, ...)
@@ -250,9 +250,27 @@ void fdprintf(const char *format, ...)
 void fdlogprintf(const char *format, ...)
 {
     va_list vl;
+    int prio = FDPP_PRINT_LOG;
 
+    if (strcmp("%s", format) == 0) {    // The string has been preformatted and we
+        const char *s;                  // may need to move an @ from args to format
+
+        va_start(vl, format);
+        s = va_arg(vl, const char *);
+        va_end(vl);
+
+        if (s[0] == '@') {
+            fdlogprintf("@%s", s + 1);  // call ourselves again
+            return;
+        }
+    }
+
+    if (format[0] == '@') {
+        prio |= FDPP_PRINT_LOG_NOPREFIX;
+        format++;
+    }
     va_start(vl, format);
-    fdlogvprintf(format, vl);
+    fdlogvprintf(prio, format, vl);
     va_end(vl);
 }
 
--- a/kernel/newstuff.c
+++ b/kernel/newstuff.c
@@ -106,10 +106,6 @@ long DosMkTmp(char FAR * pathname, UWORD attr)
   return rc;
 }
 
-#ifdef DEBUG
-#define DEBUG_TRUENAME
-#endif
-
 #define drLetterToNr(dr) ((unsigned char)((dr) - 'A'))
 /* Convert an uppercased drive letter into the drive index */
 #define drNrToLetter(dr) ((dr) + 'A')
--- a/kernel/newstuff.c
+++ b/kernel/newstuff.c
@@ -321,7 +321,7 @@ COUNT truename(__XFAR(const char) src, char FAR *dest, COUNT mode)
 
   fmemcpy(&TempCDS, cdsEntry, sizeof(struct cds));
   tn_printf(("CDS entry: #%u @%P (%u) '%s'\n", result, GET_FP32(cdsEntry),
-            TempCDS.cdsBackslashOffset, GET_FP32(TempCDS.cdsCurrentPath)));
+            TempCDS.cdsBackslashOffset, GET_PTR(TempCDS.cdsCurrentPath)));
   /* is the current_ldt thing necessary for compatibly??
      -- 2001/09/03 ska*/
   current_ldt = cdsEntry;
@@ -429,7 +429,8 @@ COUNT truename(__XFAR(const char) src, char FAR *dest, COUNT mode)
 
     if (!(mode & CDS_MODE_SKIP_PHYSICAL))
     {
-      tn_printf(("SUBSTing from: %s\n", cp));
+      tn_printf(("SUBSTing from: %s\n", GET_PTR(cp)));
+
 /* What to do now: the logical drive letter will be replaced by the hidden
    portion of the associated path. This is necessary for NETWORK and
    SUBST drives. For local drives it should not harm.
--- a/kernel/nls.c
+++ b/kernel/nls.c
@@ -34,6 +34,7 @@
  *		==ska*/
 
 #include "portab.h"
+#include "debug.h"
 #include "globals.h"
 #include "pcb.h"
 #include "nls.h"
--- a/kernel/nls.c
+++ b/kernel/nls.c
@@ -261,11 +261,10 @@ STATIC VOID upMMem(UBYTE FAR * map, UBYTE FAR * str, unsigned len)
 
   oldStr = str;
   oldLen = len;
-  log(("NLS: upMMem(): len=%u, %04x:%04x=\"", len, FP_SEG(str),
-       FP_OFF(str)));
+  DebugPrintf(("NLS: upMMem(): len=%u, %P=\"", len, GET_FP32(str)));
   for (c = 0; c < len; ++c)
-    DebugPrintf(("%c", str[c] > 32 ? str[c] : '.'));
-  DebugPrintf(("\"\n"));
+    DebugPrintf(("@%c", str[c] > 32 ? str[c] : '.'));
+  DebugPrintf(("@\"\n"));
 #endif
   if (len)
     do
@@ -280,8 +279,8 @@ STATIC VOID upMMem(UBYTE FAR * map, UBYTE FAR * str, unsigned len)
 #ifdef NLS_DEBUG
   DebugPrintf(("NLS: upMMem(): result=\""));
   for (c = 0; c < oldLen; ++c)
-    DebugPrintf(("%c", oldStr[c] > 32 ? oldStr[c] : '.'));
-  DebugPrintf(("\"\n"));
+    DebugPrintf(("@%c", oldStr[c] > 32 ? oldStr[c] : '.'));
+  DebugPrintf(("@\"\n"));
 #endif
 }
 
@@ -523,11 +522,10 @@ VOID DosUpFMem(VOID FAR * str, unsigned len)
 {
 #ifdef NLS_DEBUG
   unsigned c;
-  log(("NLS: DosUpFMem(): len=%u, %04x:%04x=\"", len, FP_SEG(str),
-       FP_OFF(str)));
+  DebugPrintf(("NLS: DosUpFMem(): len=%u, %P=\"", len, GET_FP32(str)));
   for (c = 0; c < len; ++c)
-    DebugPrintf(("%c", ((char FAR *)str)[c] > 32 ? ((char FAR *)str)[c] : '.'));
-  DebugPrintf(("\"\n"));
+    DebugPrintf(("@%c", ((char FAR *)str)[c] > 32 ? ((char FAR *)str)[c] : '.'));
+  DebugPrintf(("@\"\n"));
 #endif
   if (nlsInfo.actPkg->flags & NLS_FLAG_DIRECT_FUPCASE)
     nlsFUpMem(nlsInfo.actPkg, str, len);
@@ -698,10 +696,10 @@ UWORD ASMCFUNC syscall_MUX14(iregs FAR * regs)
       {
         unsigned j;
         BYTE FAR *p;
-        log(("NLS: MUX14(FILE_UPMEM): len=%u, %04x:%04x=\"", regs->CX, regs->ES, regs->DI));
+        DebugPrintf(("NLS: MUX14(FILE_UPMEM): len=%u, %04x:%04x=\"", regs->CX, regs->ES, regs->DI));
         for (j = 0, p = MK_FP(regs->ES, regs->DI); j < regs->CX; ++j)
-          DebugPrintf(("%c", p[j] > 32 ? p[j] : '.'));
-        DebugPrintf(("\"\n"));
+          DebugPrintf(("@%c", p[j] > 32 ? p[j] : '.'));
+        DebugPrintf(("@\"\n"));
       }
 #endif
       nlsFUpMem(nls, MK_FP(regs->ES, regs->DI), regs->CX);
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -35,4 +35,13 @@ jobs:
       run: ./ci_setup_test.sh
 
     - name: Test
+      id: test
       run: ./ci_test.sh
+
+    - name: Upload failure logs
+      if: ${{ always() && (steps.test.outcome == 'failure') }}
+      uses: actions/upload-artifact@v6
+      with:
+        name: failure-logs
+        path: ${{ env.DIR_TEST_DOSEMU }}/test_*
+
--- a/hdr/debug.h
+++ b/hdr/debug.h
@@ -48,6 +48,30 @@ void logprintf(const char *format, ...) PRINTF(1);
 
 /* use to limit output to debug builds */
 #ifdef DEBUG
+
+/* enable or disable various chunks of debug output */
+
+/* show stored IRQ vectors */
+#define DEBUGIRQ
+/* display output during kernel config processing phase */
+#define DEBUGCFG
+/* display info on various DOS functions (dosfns.c) */
+#define DEBUGDOSFNS
+/* extra debug output related to chdir */
+#define CHDIR_DEBUG
+/* extra debug output related to findfirst */
+#define FIND_DEBUG
+/* display info on various DOS directory functions (fatdir.c) */
+#define DEBUGFATDIR
+/* extra output during read/write operations */
+#define DSK_DEBUG
+/* display info on various FAT handling functions (fatfs.c) */
+#define DEBUGFATFS
+/* debug truename */
+#define DEBUG_TRUENAME
+/* display nls conversion */
+#define NLS_DEBUG
+
 #ifdef DEBUG_PRINT_COMPORT
 #define DebugPrintf(x) dbgc_printf x
 #else
@@ -61,12 +85,6 @@ void logprintf(const char *format, ...) PRINTF(1);
    keep around for later use.  */
 #define DDebugPrintf(x)
 
-
-/* enable or disable various chunks of debug output */
-
-/* show stored IRQ vectors */
-/* #define DEBUGIRQ */
-
 /* show output related to moving kernel into HMA */
 #ifdef DEBUG
 #define HMAInitPrintf(x) DebugPrintf(x)
@@ -74,30 +92,18 @@ void logprintf(const char *format, ...) PRINTF(1);
 #define HMAInitPrintf(x)
 #endif
 
-/* display output during kernel config processing phase */
-/* #define DEBUGCFG */
 #ifdef DEBUGCFG
 #define CfgDbgPrintf(x) DebugPrintf(x)
 #else
 #define CfgDbgPrintf(x)
 #endif
 
-/* display info on various DOS functions (dosfns.c) */
-/* #define DEBUGDOSFNS */
 #ifdef DEBUGDOSFNS
 #define DFnsDbgPrintf(x) DebugPrintf(x)
 #else
 #define DFnsDbgPrintf(x)
 #endif
 
-/* extra debug output related to chdir */
-/* #define CHDIR_DEBUG */
-
-/* extra debug output related to findfirst */
-/* #define FIND_DEBUG */
-
-/* display info on various DOS directory functions (fatdir.c) */
-/* #define DEBUGFATDIR */
 #ifdef DEBUGFATDIR
 #define FDirDbgPrintf(x) DebugPrintf(x)
 #else
@@ -107,25 +113,17 @@ void logprintf(const char *format, ...) PRINTF(1);
 /* extra debug output when transferring I/O chunks of data */
 /* #define DISPLAY_GETBLOCK */
 
-/* extra output during read/write operations */
-/* #define DSK_DEBUG */
-
-/* display info on various FAT handling functions (fatfs.c) */
-/* #define DEBUGFATFS */
 #ifdef DEBUGFATFS
 #define FatFSDbgPrintf(x) DebugPrintf(x)
 #else
 #define FatFSDbgPrintf(x)
 #endif
 
-/* debug truename */
-/* #define DEBUG_TRUENAME */
 #ifdef DEBUG_TRUENAME
 #define tn_printf(x) DebugPrintf(x)
 #else
 #define tn_printf(x)
 #endif
-
 
 /* ensure printf is prototyped */
 #if defined(DEBUG) || defined(DEBUGIRQ) || defined(DEBUGCFG) || \
--- a/include/fdpp/thunks.h
+++ b/include/fdpp/thunks.h
@@ -43,6 +43,7 @@ int FdppCall(struct vm86_regs *regs);
 int FdppCtrl(int idx, struct vm86_regs *regs);
 
 enum { FDPP_PRINT_LOG, FDPP_PRINT_TERMINAL, FDPP_PRINT_SCREEN };
+#define FDPP_PRINT_LOG_NOPREFIX (1 << 16)
 enum { ASM_CALL_OK, ASM_CALL_ABORT };
 enum { FDPP_RET_ABORT = -1, FDPP_RET_OK, FDPP_RET_NORET };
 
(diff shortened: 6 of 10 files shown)


kernel/nls.c
```diff
@@ -257,14 +257,17 @@ STATIC VOID upMMem(UBYTE FAR * map, UBYTE FAR * str, unsigned len)
...
   unsigned oldLen;
+  char buf[1024];
+  int n;
 
   oldStr = str;
   oldLen = len;
-  log(("NLS: upMMem(): len=%u, %04x:%04x=\"", len, FP_SEG(str),
-       FP_OFF(str)));
+
+  n = snprintf(buf, sizeof(buf), "NLS: upMMem(): len=%u, %04x:%04x=\"", len, FP_SEG(str), FP_OFF(str));
```
stsp: You can use %P here?
IIRC it prints the same seg:off
pair.

kernel/nls.c
```diff
@@ -257,14 +257,17 @@ STATIC VOID upMMem(UBYTE FAR * map, UBYTE FAR * str, unsigned len)
...
   for (c = 0; c < len; ++c)
-    DebugPrintf(("%c", str[c] > 32 ? str[c] : '.'));
-  DebugPrintf(("\"\n"));
+    n += snprintf(buf + n, sizeof(buf) - n, "%c", str[c] > 32 ? str[c] : '.');
```
stsp: It is possible to add some
silencer, eg `@` symbol, to
avoid prefixing. Then you
can leave the old logic.

kernel/newstuff.c
```diff
@@ -106,10 +106,6 @@ long DosMkTmp(char FAR * pathname, UWORD attr)
   return rc;
 }
 
-#ifdef DEBUG
-#define DEBUG_TRUENAME
-#endif
```
stsp, line 111: So what does this fix, is the
macro still undefined? Maybe
move it before the inclusion
of debug.h? Or move into debug.h?

kernel/nls.c
andrewbird: Sounds good.

kernel/nls.c
andrewbird: So the '@' handling is already implemented somewhere, or I have to add some test in dosemu's fdpp_print() to make the prepend of 'fdpp: ' optional?

stsp: I think it is not implemented
for fdpp right now, but it is
implemented for `error()` in
dosemu, so I suppose its better
to use the same logic.

kernel/newstuff.c
andrewbird: The macro is the fine tune switch of whether to print debug messages from the truename code. There are others DEBUG_FIND, CHDIR_DEBUG etc. None of the others are set locally, they are either uncommented in debug.h or you can -DDEBUG_TRUENAME.

kernel/nls.c
andrewbird: Cool, I'll do that, it will look nicer.

kernel/newstuff.c
stsp: Ok but can you make sure that
once DEBUG is defined, they all
are ON in debug.h?
We do not use things like
-DDEBUG_TRUENAME currently,
so your fix is insufficient.

andrewbird: That suits me.

fdpp/thunks.cc
```diff
@@ -235,7 +235,10 @@ void fdvprintf(const char *format, va_list vl)
 
 static void fdlogvprintf(const char *format, va_list vl)
 {
-    fdpp->print(FDPP_PRINT_LOG, format, vl);
+    if (format[0] == '@')
+        fdpp->print(FDPP_PRINT_LOG_NOPREFIX, format + 1, vl);
```
stsp: Just add flag, shift it to
16 to not clash with the value.

andrewbird:
I have a crash in truename at adjust_far() once I enable the debugging. The trigger is a super long argument (128 chars) which actually was unnecessary so I have removed it in my dosemu branch. Not sure whether it's important, after all it's a ridiculous size,  but I thought I'd mention it.

fdpp/thunks.cc
```diff
@@ -250,9 +250,27 @@ void fdprintf(const char *format, ...)
 void fdlogprintf(const char *format, ...)
 {
     va_list vl;
+    int prio = FDPP_PRINT_LOG;
+
+    if (strcmp("%s", format) == 0) {    // The string has been preformatted and we
```
stsp, line 255: I think it would be reasonable
to demand `@` in a format string
itself. I'd like to avoid va_arg() here.

stsp:
`*** stack smashing detected ***: terminated`
Wow! Can you provide a minimal
reproducer? I suppose maybe some
buffer overflow on vsprintf(), so maybe
you can also find it yourself. :)

andrewbird:
> I suppose maybe some
buffer overflow on vsprintf(), so maybe
you can also find it yourself. :)

I already looked for a couple of hours, but my lack of C++ knowledge stops me from finding anything other than it occurs in the call `adjust_far()` in truename.

> Wow! Can you provide a minimal
reproducer?

1/ Yes, use dosemu devel (still has the excessively long and unnecessary argument in the batch file `testit.bat`), with fdpp of this branch at c71328b886b563ed4a18642bd82b65a2cfaf72ad (which is compilation fixes, and turning on of all DEBUG options, but none of my '@' shenanigans).
2/ Setup test as
 `env NO_TESTRUN=1 test/test_dosemu.py PPDOSGITTestCase.test_mfs_sfn_file_read`.
3/ Run dosemu as
Start dosemu as whatever the above printed, but in my case I renamed the log file
`bin/dosemu -f /tmp/test-imagedir/dosemu.conf -n -q -o crash.log -td --Fimagedir /tmp/test-imagedir`
4/ Run test batchfile as is, but note the crash only occurs with `set LFN=n`
`C:\>testit`

Here's mine.
~~~
ajb@calypso:/clients/common/dosemu2.git$ bin/dosemu -f /tmp/test-imagedir/dosemu.conf -n -q -o crash.log -td --Fimagedir /tmp/test-imagedir
FDPP kernel 1.7 [GIT: 1.10-66-gc71328b8]
Kernel compatibility 7.10 - clang - FAT32 support

Written by @stsp, FDPP project.
Based on FreeDOS sources (C) Pasquale J. Villani and The FreeDOS Project.

This program is free software: you can redistribute it and/or modify
it under the terms of the GNU General Public License as published by
the Free Software Foundation, either version 3 of the License, or
(at your option) any later version.

C: HD1, Pri[ 1], CHS=    0-1-1, start=     0 MB, size=    10 MB
D: HD2, Pri[ 1], CHS=    0-1-1, start=     0 MB, size=    10 MB
E: HD3, Pri[ 1], CHS=    0-1-1, start=     0 MB, size=  2000 MB
F: HD4, Pri[ 1], CHS=    0-1-1, start=     0 MB, size=  2000 MB
G: HD5, Pri[ 1], CHS=    0-1-1, start=     0 MB, size=  2000 MB
EMUFS host file and print access, non-resident driver mode                     
dosemu XMS 3.0 & UMB support enabled
dosemu EMS driver rev 0.9 installed.
dosemu CDROM driver installed (V0.2)
Process 0 starting: F:\command.com /e:384 /p /k %FDPP_AUTOEXEC% /M2
"Warning: USERDRV not set"

FreeDOS userspace tools can be installed with: insfdusr

BLASTER=A220 I5 D1 H5 P300 T6
MIDI=SYNTH:2 MAP:E MODE:0
Welcome to dosemu2!
    Build 2.0pre9-dev-20260706-4798-g5811247aa
Interactive Prompt!
C:\>testit
testit
C:\>set LFN=n
C:\>d:
D:\>c:\mfsread shrtname.txt FfaaE90dDC39F9C30Db4Ea75EbA420B1cB1DaEaedB89Cf51E3FA4B461878290EeCC0f42cd6e96fCBD6c66350edCc3054e04eB177eF1CDF7bf2D4e45D36225D28
Segmentation fault (Address not mapped to object [0x727e000a293a])
Terminated with signal 11
~~~

stsp:
```
This commit does not belong to any branch on this repository, and may belong to a fork outside
of the repository.
```
to start with.

Why is this a minimal reproducer?
Minimal reproducer is:
- change this file this way
- run dosemu as usual
- observe its crashing

stsp:
I suppose you meant this link:
https://github.com/andrewbird/fdpp/commit/c71328b886b563ed4a18642bd82b65a2cfaf72ad

stsp:
```
$ bin/dosemu -f /tmp/test-imagedir/dosemu.conf -n -q -o crash.log -td --Fimagedir /tmp/test-imagedir
ERROR: --Fimagedir: /tmp/test-imagedir does not exist
ERROR: /tmp/test-imagedir/dosemu.conf is missing
```

andrewbird:
You didn't set up the like the instructions said
`env NO_TESTRUN=1 test/test_dosemu.py PPDOSGITTestCase.test_mfs_sfn_file_read`

andrewbird:
or more likely you did but didn't use the command it printed (what you posted looks like mine)

stsp:
The problem is that _vsnprintf()
applies size to `fmt` argument,
i.e. it counts chars consumed,
not chars outputted. So `%s` is
counted like 1 char, which results
in overflow. Given that this code
is a pure crap, it uses a lot of global
vars and small buffers like
`BYTE s[11]`, I don't see a quick fix.
Maybe the only reasonable thing
to do is to port from musl, but
you'll need to add a few things
there, like `%P`.
So will you work on that?

andrewbird:
Are you set on musl it looked quite complicated to extract the salient part? I looked for an alternative, this popped up https://github.com/nothings/stb/blob/master/stb_sprintf.h license is MIT or Public Domain.

stsp:
I am not set for musl by
any means, but public domain?
I simply don't know.
If you show me some document
which says its fine for gpl projects -
then likely ok...

andrewbird:
No, it's Public Domain **OR** MIT, the user gets to choose which he wants to follow.

stsp:
That is written in the end, and
not in the beginning...
Anyway, I haven't heard anything.
I'll just see MIT in your patch
(at the beginning!), and fare well.

andrewbird:
https://github.com/nothings/stb/tree/master#faq

stsp:
Ok, choose MIT and remove
the rest then.

andrewbird:
Will the new vsnprintf need to be in the kernel C code or in the C++ part?

stsp:
Just replace the buggy prf.c.

andrewbird:
Can you tell me what implementation of put_console() is used for fdpp, it's all ordered and not obvious which was built in the old pdf.?. also do I need to implement hexd, there's only one use of it outside of per.c?

stsp:
As surprising as it may seem,
not the one under `#ifdef DOSEMU`. :)
```
  iregs r = {};
  r.a.b.l = c;
  call_intr(0x29, MK_FAR_SCP(r));
```
I am pretty sure this one: MK_FAR_SCP()
is what only fdpp does.

hexd() - just copy/paste it?
It doesn't need in the same file
with printf's.

andrewbird:
I made a draft PR that seems to work #300, but I wonder if there's some way that _sprintf() could do the GET_PTR/GET_FP32 translation instead of them being splattered all over the source?

stsp:
Possibly, but not in this PR, since
currently they are already "splattered".
I dont think it is possible to do so
w/o either adding new formats
(unwanted) or do some nasty mods
in the generic printf code, but I didn't
think of that a lot. Try to find a
minimally invasive approach, this
will require some creativity.

Четверг, 9 июля 2026 г получено от Andrew Bird:

stsp:
I think actually this can be done by
adding just 1 new conversion for all
fdpp types, eg %A (auto) that will
extract the type and value automatically.
But that will still mean changes to
the existing callers, which doesn't
make it any better than not doing
anything.

Четверг, 9 июля 2026 г получено от Andrew Bird:

andrewbird:
Google gave me some hints on what to experiment with to implement GET_PTR() / GET_FP32() conversions outside of the kernel (for a later PR if successful). If I want to implement _sprintf() etc as C++ using templates and concepts I guess I'll need to put it in a separate .cpp file somewhere. Would that live in fdpp directory alongside farptr.*?

stsp:
Yes, you can put it into fdpp,
but I don't think this strategy
(unpacking automatically, then
pass) will ever work. Try it, but
I am very sceptical.

andrewbird:
I don't think it will work either, certainly not in a portable fashion.

stsp:
Hmm... Actually there is a way.
You just need to extend the mkfar.sh
script with this:
https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-std-print.html

stsp:
And I really think that's worth doing,
as that will get us closer to the better
gcc port. gcc is weak on formats, so
we can use that helper.

andrewbird:
My clang-tidy (v18) on Noble doesn't seem to have that option, so I guess it can't be part of the build yet.

stsp:
Are you sure?
AI says it needs clang-17, and
is available in Noble.

andrewbird:
I was expecting to see the option when the usage as displayed. However doing `clang-tidy --list-checks -checks=\*` does show it.

andrewbird:
I woke up this morning and wondered if we could process the Dos C files with coccinelle to do the replacements. But to be honest I don't really understand the syntax and was pretty sure it couldn't handle variadic replacements. Any way Google's suggestion was to process it with python using tree sitter. That turned out to be non starter, but plain python seems to work okay. How do you feel about AI generated stuff? I have it converting a file `dosfns.c` lifted straight from FreeDOS kernel to something that looks like it did the right thing. I have a deep feeling of dissatisfaction, while the tool does what I asked, I don't feel like I wrote it, only guided the AI. So where's the fun in that?

Anyway, here's a sample of its output

```diff
$ python3 fix_printf10.py mytest.c mytest.cpp


$ diff -surN mytest.c mytest.cpp 
--- mytest.c	2026-07-10 10:18:51.408344281 +0100
+++ mytest.cpp	2026-07-10 10:41:53.024348288 +0100
@@ -539,19 +539,19 @@
   sftp->sft_attrib = attrib = attrib | D_ARCHIVE;
 
   /* check for a (local) device */
-  DebugPrintf(("result for %Fs is %s%s%s%s",fname,(result & IS_DEVICE)?"device ":"X",(result & IS_NETWORK)?"network ":"N",(dhp = IsDevice(fname)) != NULL?"IsDevice!=NULL ":"="));
+  DebugPrintf(("result for %s is %s%s%s%s",GET_PTR(fname),(result & IS_DEVICE)?"device ":"X",(result & IS_NETWORK)?"network ":"N",(dhp = IsDevice(fname)) != NULL?"IsDevice!=NULL ":"="));
 
   /* same line just different formatting */
-  DebugPrintf(("result for %Fs is %s%s%s%s",
-        fname,
+  DebugPrintf(("result for %s is %s%s%s%s",
+        GET_PTR(fname),
         (result & IS_DEVICE)?"device ":"X",
         (result & IS_NETWORK)?"network ":"N",
         (dhp = IsDevice(fname)) != NULL?"IsDevice!=NULL ":"="));
 
   /* another line with %P */
-  DebugPrintf(("result for %Fs is %P\n",
-        fname,
-        fname));
+  DebugPrintf(("result for %s is %P\n",
+        GET_PTR(fname),
+        GET_FP32(fname)));
 
   if ((result & IS_DEVICE) && !(result & IS_NETWORK) && (dhp = IsDevice(fname)) != NULL)
   {
@@ -1004,7 +1004,7 @@
   COUNT result;
 
 #if defined(CHDIR_DEBUG)
-  DebugPrintf(("DosChangeDir: %Fs\n", s));
+  DebugPrintf(("DosChangeDir: %s\n", GET_PTR(s)));
 #endif
   result = truename(s, PriPathName, CDS_MODE_CHECK_DEV_PATH);
   if (result < SUCCESS)
@@ -1022,14 +1022,14 @@
     return DE_PATHNOTFND;
 
 #if defined(CHDIR_DEBUG)
-  DebugPrintf(("Remote Chdir: n='%Fs' p='%s\n", s, PriPathName));
+  DebugPrintf(("Remote Chdir: n='%s' p='%s\n", GET_PTR(s), PriPathName));
 #endif
   /* now get fs to change to new          */
   /* directory                            */
   result = (result & IS_NETWORK ? network_redirector(REM_CHDIR) :
             dos_cd(PriPathName));
 #if defined(CHDIR_DEBUG)
-  DebugPrintf(("status = %04x, new_path='%Fs'\n", result, current_ldt->cdsCurrentPath));
+  DebugPrintf(("status = %04x, new_path='%s'\n", result, GET_PTR(current_ldt->cdsCurrentPath)));
 #endif
   /* on error result should be negative, on success for network call value
      should be ignored as some redirectors e.g. Lantastic return count on success */
@@ -1071,7 +1071,7 @@
   int rc;
   register dmatch FAR *dmp = dta;
 
-  DebugPrintf(("DosFindFirst(%Fs, 0x%04x)\n", name, attr));
+  DebugPrintf(("DosFindFirst(%s, 0x%04x)\n", GET_PTR(name), attr));
   rc = truename(name, PriPathName,
                 CDS_MODE_CHECK_DEV_PATH | CDS_MODE_ALLOW_WILDCARDS);
   if (rc < SUCCESS)
@@ -1090,7 +1090,7 @@
   SAttr = (BYTE) attr;
 
 #if defined(FIND_DEBUG)
-  DebugPrintf(("Remote Find: n='%Fs\n", PriPathName));
+  DebugPrintf(("Remote Find: n='%s\n", GET_PTR(PriPathName)));
 #endif
 
   dta = &sda_tmp_dm;
```

And the python
```python
import sys
import re

# Combined transformation configuration map
TRANSFORMS_MAP = {
    'Fs': {
        'wrapper': 'GET_PTR',
        'replacement': 's'
    },
    'P': {
        'wrapper': 'GET_FP32',
        'replacement': None  # Change to a string if you need to mutate %P -> %p
    }
}

# Borland + Standard Function Index Map
FORMAT_INDEX_MAP = {
    'printf': 0, 'fprintf': 1, 'sprintf': 1, 'snprintf': 2,
    'cprintf': 0, 'wprintf': 0, 'fwprintf': 1, 'swprintf': 1,
    'DebugPrintf': 0
}

FUNC_PATTERN = re.compile(rf"\b({'|'.join(FORMAT_INDEX_MAP.keys())})\s*\(", re.MULTILINE)

def extract_balanced_args_with_positions(text, start_pos):
    """
    Steps through characters starting at an open parenthesis.
    Returns a list of tuples containing: (arg_text, global_start_idx, global_end_idx).
    Preserves all internal whitespace, tabs, and newlines exactly.
    """
    args = []
    current_arg = []
    paren_depth = 0
    in_string = False
    escape = False
    
    arg_start = start_pos + 1
    i = start_pos
    
    while i < len(text):
        char = text[i]
        
        if in_string:
            current_arg.append(char)
            if escape:
                escape = False
            elif char == '\\':
                escape = True
            elif char == '"':
                in_string = False
        else:
            if char == '"':
                in_string = True
                current_arg.append(char)
            elif char == '(':
                paren_depth += 1
                if paren_depth > 1:
                    current_arg.append(char)
            elif char == ')':
                paren_depth -= 1
                if paren_depth == 0:
                    if current_arg or len(args) > 0:
                        args.append(("".join(current_arg), arg_start, i))
                    return i + 1, args
                current_arg.append(char)
            elif char == ',' and paren_depth == 1:
                args.append(("".join(current_arg), arg_start, i))
                current_arg = []
                arg_start = i + 1
            else:
                current_arg.append(char)
        i += 1
    return -1, []

def process_file(input_path, output_path):
    with open(input_path, 'r', encoding='utf-8', errors='ignore') as f:
        source_code = f.read()

    has_errors = False
    modifications = []
    pos = 0
    
    while True:
        match = FUNC_PATTERN.search(source_code, pos)
        if not match:
            break
            
        func_name = match.group(1)
        func_start = match.start()
        open_paren_pos = match.end() - 1
        
        end_pos, args = extract_balanced_args_with_positions(source_code, open_paren_pos)
        
        if end_pos == -1 or not args:
            pos = match.end()
            continue

        # Double-parentheses macro unpacking layer
        if len(args) == 1 and args[0][0].strip().startswith('(') and args[0][0].strip().endswith(')'):
            inner_text_raw, inner_start, inner_end = args[0]
            first_paren = inner_text_raw.find('(')
            _, inner_args = extract_balanced_args_with_positions(inner_text_raw, first_paren)
            if inner_args:
                args = [(txt, inner_start + s, inner_start + e) for txt, s, e in inner_args]

        fmt_index = FORMAT_INDEX_MAP[func_name]
        if fmt_index >= len(args):
            pos = end_pos
            continue

        fmt_arg_text, fmt_start, fmt_end = args[fmt_index]
        fmt_arg_stripped = fmt_arg_text.strip()
        
        if not (fmt_arg_stripped.startswith('"') and fmt_arg_stripped.endswith('"')):
            pos = end_pos
            continue

        fmt_str = fmt_arg_stripped[1:-1]
        
        specifier_pattern = re.compile(r'%(?:%|[0-9.+\-*#lhzj]*[a-zA-Z]+)')
        specifiers = [t for t in specifier_pattern.findall(fmt_str) if t != '%%']
        vargs = args[fmt_index + 1:]

        updated_fmt_str = fmt_str
        format_string_modified = False

        for i, spec in enumerate(specifiers):
            if i >= len(vargs):
                break
                
            arg_text, arg_start, arg_end = vargs[i]
            arg_stripped = arg_text.strip()
            
            # Identify if this specifier matches any configured token keys
            target_token = next((token for token in TRANSFORMS_MAP if token in spec), None)
            
            if target_token:
                # ENFORCED POLICY ERROR: Fail build if un-casted macros or functions sit in the target area
                if '(' in arg_stripped and not re.match(r'^\s*\([^)]+\)\s*[a-zA-Z_]', arg_stripped):
                    line_no = source_code.count('\n', 0, func_start) + 1
                    print(f"Error: [{input_path}:{line_no}] Found restricted macro/function call '{arg_stripped}' inside {func_name}.", file=sys.stderr)
                    has_errors = True
                    continue

                transform_config = TRANSFORMS_MAP[target_token]
                wrapper = transform_config['wrapper']
                replacement = transform_config['replacement']

                # Wrap the text chunk while safely preserving leading and trailing spaces
                leading_spaces = arg_text[:len(arg_text)-len(arg_text.lstrip())]
                trailing_spaces = arg_text[len(arg_text.rstrip()):]
                
                wrapped_arg_text = f"{leading_spaces}{wrapper}({arg_stripped}){trailing_spaces}"
                modifications.append((arg_start, arg_end, wrapped_arg_text))
                
                # Check if the string literal specifier itself needs editing
                if replacement is not None:
                    replaced_spec = spec.replace(target_token, replacement)
                    updated_fmt_str = updated_fmt_str.replace(spec, replaced_spec, 1)
                    format_string_modified = True

        if format_string_modified and not has_errors:
            leading_fmt_spaces = fmt_arg_text[:len(fmt_arg_text)-len(fmt_arg_text.lstrip())]
            trailing_fmt_spaces = fmt_arg_text[len(fmt_arg_text.rstrip()):]
            
            new_fmt_arg_text = f'{leading_fmt_spaces}"{updated_fmt_str}"{trailing_fmt_spaces}'
            modifications.append((fmt_start, fmt_end, new_fmt_arg_text))
            
        pos = end_pos

    if has_errors:
        sys.exit(1)

    modifications.sort(key=lambda x: x[0], reverse=True)
    
    mutable_source = list(source_code)
    for start, end, new_text in modifications:
        mutable_source[start:end] = list(new_text)

    with open(output_path, 'w', encoding='utf-8') as f:
        f.write("".join(mutable_source))

if __name__ == "__main__":
    if len(sys.argv) < 3:
        print("Usage: python fix_printf.py <input.c> <output.cpp>")
        sys.exit(1)
    process_file(sys.argv[1], sys.argv[2])

```

stsp:
So why not to use clang-tidy, why
something else was needed? Anyway,
if you can review your AI code, then
I don't know why it can't be applied.

stsp:
... why it can't be applied, correction.

andrewbird:
Sure, it doesn't look particularly frightening to review, and it looks like it will achieve the goal. I know I don't write the toolchain tools, so I shouldn't be bothered, but AI gifted code is unsatisfying especially now it's getting better. I guess I should stop asking it questions...

I'll investigate clang-tidy though first.

stsp:
> but AI gifted code is unsatisfying especially now it's getting better. I guess I should stop asking it
> questions...

I don't understand what you say.
If AI helps solving the task quickly
(you of course need to add the time
needed to review its code), then
you need to use it. For me, it helps
a lot to just discuss things with AI.
So instead of asking it "how", you
can try asking "why not this way".
This is similar to code review. When
you open some PR, you basically also
asking the reviewer "why not this way".

andrewbird:
Yes, asking AI questions is how these things get started for me.  If I were talking to a human reviewer I wouldn't expect or desire for an initial code snippet to get fleshed out into a working solution. I need to learn to stop asking follow up questions, and if that code snippet doesn't compile, just accept it, figure out what and fix it myself.

I think it can be laziness on my part, what the AI doesn't understand is that for me the goal is not the fix but the process. Handing me a turnkey solution may solve the task at hand, but is not enjoyable for me and in the end the goal was not met.

I know this is coming across as an ungrateful rant, I'm sorry that's the case, and I'm sure it's a 'me' problem.

stsp:
> Handing me a turnkey solution may solve the task at hand, but is not enjoyable for me and in the
> end the goal was not met.

The problem with software development,
is that this is a resource-hungry process.
Aim to do everything yourself - doesn't
work. If you think it did something for
you and you have nothing to do - is wrong.
Quite the opposite, you can do 100 times
more by off-loading things to AI. I tried
to off-load some coding to AI, but so far
the results were unsatisfactory, so I ended
up doing that myself, which is slow. If you
see it already suits your needs - use it!
And you won't get out of business, quite
the opposite.

PR #300, opened 2026-07-09
Printf 01

andrewbird:
First draft of using a printf replacement, and added enhancements to support  %P ~and %Fs~.

### DIFF
--- a/kernel/stb_sprintf.h
+++ b/kernel/stb_sprintf.h
@@ -102,8 +102,8 @@ int stbsp_vsprintfcb( STBSP_SPRINTFCB * callback, void * user, char * buf, char 
   The buffer you pass in must hold at least STB_SPRINTF_MIN characters.
     // you return the next buffer to use or 0 to stop converting
 
-void stbsp_set_separators( char comma, char period )
-  Set the comma and period characters to use.
+void stbsp_set_separators( char colon, char comma, char period )
+  Set the colon, comma and period characters to use.
 
 FLOATS/DOUBLES:
 ===============
@@ -233,7 +233,7 @@ STBSP__PUBLICDEC int STB_SPRINTF_DECORATE(sprintf)(char *buf, char const *fmt, .
 STBSP__PUBLICDEC int STB_SPRINTF_DECORATE(snprintf)(char *buf, int count, char const *fmt, ...) STBSP__ATTRIBUTE_FORMAT(3,4);
 
 STBSP__PUBLICDEC int STB_SPRINTF_DECORATE(vsprintfcb)(STBSP_SPRINTFCB *callback, void *user, char *buf, char const *fmt, va_list va);
-STBSP__PUBLICDEC void STB_SPRINTF_DECORATE(set_separators)(char comma, char period);
+STBSP__PUBLICDEC void STB_SPRINTF_DECORATE(set_separators)(char colon, char comma, char period);
 
 #endif // STB_SPRINTF_H_INCLUDE
 
@@ -280,6 +280,7 @@ static stbsp__int32 stbsp__real_to_parts(stbsp__int64 *bits, stbsp__int32 *expo,
 
 static char stbsp__period = '.';
 static char stbsp__comma = ',';
+static char stbsp__colon = ':';
 static struct
 {
    short temp; // force next field to be 2-byte aligned
@@ -293,10 +294,11 @@ static struct
    "75767778798081828384858687888990919293949596979899"
 };
 
-STBSP__PUBLICDEF void STB_SPRINTF_DECORATE(set_separators)(char pcomma, char pperiod)
+STBSP__PUBLICDEF void STB_SPRINTF_DECORATE(set_separators)(char pcolon, char pcomma, char pperiod)
 {
    stbsp__period = pperiod;
    stbsp__comma = pcomma;
+   stbsp__colon = pcolon;
 }
 
 #define STBSP__LEFTJUST 1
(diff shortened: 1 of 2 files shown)


kernel/prf-old.c
```diff
@@ -0,0 +1,613 @@
...
+  }
+  va_end(arg_copy);
+}
+
+#if !defined(FORSYS) && !defined(_INIT)
+
+extern void put_string(const char *);
+extern void put_unsigned(unsigned, int, int);
+
+void hexd(const char *title, VOID FAR * v_p, COUNT numBytes)
```
stsp: Since prf.c is just a wrapper now,
why do you move hexd() away? It
could stay?

fdpp/thunks_priv.h
```diff
@@ -24,9 +24,9 @@ void *resolve_segoff_fd(struct far_s fa);
 void thunk_call_void(struct far_s fa);
 void thunk_call_void_noret(struct far_s fa);
 struct far_s lookup_far_st(const void *ptr);
-void fdprintf(const char *format, ...) PRINTF(1);
```
stsp, line 27: No, please don't!
There are no warnings with
clang, gcc is buggy, its support
is experimental.

kernel/stb_sprintf.h
```diff
@@ -609,11 +609,22 @@ STBSP__PUBLICDEF int STB_SPRINTF_DECORATE(vsprintfcb)(STBSP_SPRINTFCB *callback,
          stbsp__int32 dp;
          char const *sn;
 
+      case 'F':
+         ++f;
+         if (f[0] != 's') {
```
stsp: I am not sure I understand
why do we need %Fs? %P is
understood, but %Fs is the
same as %s?

stsp:
It would be good to split into
2 PR, although as you wish.

kernel/prf-old.c
andrewbird: The commit history is awkward. I didn't want to break the build by having two commits, one to remove the old prf.c (build is broken), one to implement a new prf.c that uses stb_vsnprintf etc. So renamed prf.c -> prf-old.c , and created a new prf.c all in the same commit. I thought that git would spot the 100% rename (it did), and that it would recognise that prf.c was wholly new (it didn't and still shows diff from the old one). So if we want a clean diff the only thing I can think of now is to call the new file something different and alter the build. Anyway hexd() is still there, but I probably changed the order in the new file.

kernel/stb_sprintf.h
andrewbird: There's one use of %Fs. Yes GET_PTR() creates a %s compatible string. I think the original prf.c supported Fs and S for FAR strings, was just trying to follow that.

kernel/prf-old.c
stsp: > So renamed prf.c -> prf-old.c , and created a new prf.c all in the same commit.

Please explain why do you need old,
if you create the new one in the same
commit? This is essentially not obvious.

kernel/stb_sprintf.h
stsp: Either remove it or add as an
alias to %s.

kernel/prf-old.c
andrewbird: I don't, it was a cunning attempt to convince git diff to recognise the new prf.c as unrelated to the old. It didn't work and I got the worst of both worlds. I'll remove and squash.

stsp: I won't mind prfnew.c or whatever.
Then you can "cheat" git.
Plain prf.c also suits.

fdpp/thunks_priv.h
andrewbird: I removed that now, but I do get warnings with clang.
~~~
dosfns.cc:1130:41: warning: format specifies type 'double' but the argument has type 'char *' [-Wformat]
 1130 |   DebugPrintf(("Remote Find: n='%Fs\n", GET_PTR(PriPathName)));
      |   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
      |                                 %s
/clients/common/fdpp.git/fdpp/farptr.hpp:851:20: note: expanded from macro 'GET_PTR'
  851 | #define GET_PTR(f) (f).get_ptr()
      |                    ^
/clients/common/fdpp.git/fdpp/../hdr/debug.h:78:34: note: expanded from macro 'DebugPrintf'
   78 | #define DebugPrintf(x) logprintf x
      |                                  ^
1 warning generated.
~~~

kernel/stb_sprintf.h
andrewbird: I'll remove, it trips up the format linting, thinks %Fs is a double.

fdpp/thunks_priv.h
stsp: Yes but %Fs must also be removed.

andrewbird: done

kernel/prf-old.c
andrewbird: I got rid of prf-old.c now. I think I may as well just leave it as prf.c as I think git doesn't care about file names, only the percentage of similarity, I don't think renaming it will help.

stsp:
LGTM.

andrewbird:
Seems to be a problem with zero padding with %P, old code prints '0090:00b7', new '90:00B7'

andrewbird:
Okay, I have a fix for the field width.

Next question I had to choose upper or lower case hex for %P. I chose upper, but it seems the old implementation was lower. Which would you like upper or lower?

stsp:
I don't mind.
So if old was lower, feel free to
mimic that.

andrewbird:
I think I'll split this PR into printf and debug parts, so just the first two patches for this PR, okay?

stsp:
Of course the split is okey.

andrewbird:
Okay, so just the two patches now for printf replacement. I fixed the field width and made the values lowercase hex. The output log compares very well with master, with only the host pointers and random FLOPPY messages differing.

stsp:
Thanks!
This is a great work that improves
reliability and security.

PR #303, opened 2026-07-28
Ci 60

andrewbird:
Switch to 26.04 and re-enable the processor testing


stsp:
Ok, I'll better try to re-build gcc-ia16.
We've seen recently that LP works -
maybe it still does...

andrewbird:
> Ok, I'll better try to re-build gcc-ia16.

Actually, Tkchia has fixed all the bugs in his upstream so we could switch to that instead?

andrewbird:
But I don't think he has any intention of adding amd64v3 packages, so apart from the benefit of you not needing to maintain it and a couple of new functions/tests, there's no real difference.

stsp:
> Actually, Tkchia has fixed all the bugs in his upstream so we could switch to that instead?

Please, if you want to do anything
with gcc-ia16 - I'll happily transfer
everything I can. Otherwise please
don't tell me about his "fixes", as I've
already seen them.

stsp:
Or do you mean he provides an
upstream builds in PPA? In this
case go ahead and switch, and
I'll just delete my PPA.

andrewbird:
Yes he has builds for Ubuntu LTS releases in PPA.

stsp:
Including 26.04?
Then what is the problem?

andrewbird:
Yes, he has 26.04.

stsp:
Great, please switch then.

andrewbird:
okay will do that once I get dosemu2 CI done on 26.04.

Next question is nasm-segelf required now or not, as it seems that meson build doesn't use it, but make still wants it?

stsp:
No, make just says its not there,
but you can remove that msg.

andrewbird:
I was mistaken, it's not `nasm-segelf` missing that's my install problem (diagnostics from `mk-build-deps`  very poor), it's actually `dj64dev` which seems to exist in two PPAs. In `stsp-0` you've updated it so the amd64v3 package exists, in `dosemu2` it hasn't been updated since jul-22. Ideally I'd like it in `dosemu2` but I'd settle for just one place whichever.

stsp:
Thanks, rebuilding.
I am also still trying to build
gcc-ia16 for you. But, as always,
that's a PITA.

stsp:
Everything is updated.

andrewbird:
Cool, great, thanks!

